PR #63, opened 2018-08-13
Convert `Outlet` from a HOC to a standard component with a render prop

### LINKED ISSUE #60: Convert `Outlet` to use a render prop to describe the output
**Enhancement**

An `Outlet` is a basic higher order component which accepts one or more widgets on creation of the outlet. 

```ts
// separate outlet component file

export const MyOutlet = Outlet({
    index: IndexWidget, 
    main: MainWidget, 
    error: ErrorWidget }, 'outlet-name'):
```

The current implementation is problematic for typings when there are multiple widgets specified as the `Outlet`'s properties becomes a union type of all the widget's properties passed to the `Outlet`. This is not correct, as you may need to pass different values for the same property depending on the match type scenario. Another side affect means that widgets could receive extra properties that were intended for a different widget.

There are at least a couple of potential options:

#### Option 1

Move from a HOC to a component that accepts a render property to describe the outlets output. This will allow consumers to explicitly pass the properties to each of the widgets and give more control to the consumer to create customised logic on which widgets to display.

```ts
// inside your widget
render() {
    return v('div', [
        w(Outlet, { 
            to: 'outlet-name', 
            renderer: (matchDetails) => {
                // match details contains the `matchType`, `params` and `queryParams`
                return w(MyWidget, { foo: 'bar', id: matchDetails.id });
            }
        })
    ]);
}
```

#### Options 2

Keep with a HOC to a component but one that accepts a render function to describe the outlets output. This will also allow consumers to explicitly pass the properties to each of the widgets and give more control to the consumer to create customised logic on which widgets to display.

```ts
interface MyOutletProperties {
    foo: string;
}

const MyOutlet = Outlet<MyOutletProperties>((properties, outletProps) => {
     if (outletProps.isError()) {
        return w(ErrorWidget, {});
     }
     if (outletProps.isExact()) {
         return w(ExactMatchWidget, {});
     }

     // return undefined when no widget is needed to be rendered based on the match type.
     return undefined;
}, 
{ outlet: 'foo' })
```

agubler:
**Type:** feature

**Description:**

This changes the `Outlet` from a higher order component to a standard component with a render property. The render property is used to display content when a route matches providing more control over the nodes to render when an Outlet's route has been matched.

Having an `Outlet` as a HOC implies that it would be shared across the application (which generally they are not), adds complexity in it's usage by having to specify another interface for the properties that the `Outlet` can accept and has more boilerplate and indirection that if the `Outlet` is a standard widget.

```ts
render() {
    return v('div', [
        w(Outlet, { 
            id: 'outlet-name', 
            renderer: (matchDetails: MatchDetails) => {
                // match details contains the `matchType`, `params` and `queryParams`
                return v(MyWidget, { foo: 'bar', id: matchDetails.id });
            }
        })
    ]);
}
```

Resolves #60 

Alternative to #24

### DIFF
--- a/src/routing/Outlet.ts
+++ b/src/routing/Outlet.ts
@@ -14,7 +14,7 @@ export interface OutletProperties {
 
 @alwaysRender()
 export class Outlet extends WidgetBase<OutletProperties> {
-	private _handle: Handle;
+	private _handle: Handle | undefined;
 	private _matched = false;
 	private _matchedParams: Params = {};
 	private _onExit?: OnExit;
@@ -25,6 +25,7 @@ export class Outlet extends WidgetBase<OutletProperties> {
 		const item = this.registry.getInjector<Router>(routerKey);
 		if (this._handle) {
 			this._handle.destroy();
+			this._handle = undefined;
 		}
 		if (item) {
 			this._handle = item.invalidator.on('invalidate', () => {
@@ -92,4 +93,5 @@ export class Outlet extends WidgetBase<OutletProperties> {
 		return null;
 	}
 }
+
 export default Outlet;
--- a/src/routing/Outlet.ts
+++ b/src/routing/Outlet.ts
@@ -1,100 +1,95 @@
-import { DNode, WidgetBaseInterface } from '../widget-core/interfaces';
+import { DNode } from '../widget-core/interfaces';
 import { WidgetBase } from '../widget-core/WidgetBase';
-import { w } from '../widget-core/d';
-import { inject } from '../widget-core/decorators/inject';
 import { alwaysRender } from '../widget-core/decorators/alwaysRender';
-import { OnEnter, Component, OutletOptions, OutletComponents, Outlet, Params, OutletContext } from './interfaces';
+import { MatchDetails, OnExit, OutletContext, OnEnter, Params } from './interfaces';
 import { Router } from './Router';
+import { diffProperty } from '../widget-core/decorators/diffProperty';
+import { Handle } from '../core/Destroyable';
 
-export function isComponent<W extends WidgetBaseInterface>(value: any): value is Component<W> {
-	return Boolean(value && (typeof value === 'string' || typeof value === 'function' || typeof value === 'symbol'));
+export interface OutletProperties {
+	renderer: (matchDetails: MatchDetails) => DNode;
+	outlet: string;
+	routerKey?: string;
 }
 
-export function getProperties(router: Router, properties: any) {
-	return { router };
-}
+@alwaysRender()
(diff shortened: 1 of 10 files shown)


src/routing/interfaces.d.ts
```diff
@@ -129,42 +123,37 @@ export interface OnExit {
...
 }
 
-/**
- * Outlet options that can be configured
- */
-export interface OutletOptions {
-	key?: RegistryLabel;
-	mapParams?: MapParams;
-}
+export interface MatchDetails {
```
agubler: Is `OutletDetails` a better name?

tomdye: I think `MatchDetails` is more appropriate as it's the details for the current match, not the outlet as a whole. Perhaps `MatchProperties` though?

src/routing/Outlet.ts
```diff
@@ -1,100 +1,97 @@
...
+import { MatchDetails, OnExit, OutletContext, OnEnter, Params } from './interfaces';
 import { Router } from './Router';
+import { diffProperty } from '../widget-core/decorators/diffProperty';
+import { Handle } from '../core/Destroyable';
 
-export function isComponent<W extends WidgetBaseInterface>(value: any): value is Component<W> {
-	return Boolean(value && (typeof value === 'string' || typeof value === 'function' || typeof value === 'symbol'));
+export interface OutletProperties {
+	renderer: (matchDetails: MatchDetails) => DNode;
+	outlet: string;
```
tomdye: perhaps rename to `id` ?

src/routing/Outlet.ts
```diff
@@ -1,100 +1,97 @@
...
-import { OnEnter, Component, OutletOptions, OutletComponents, Outlet, Params, OutletContext } from './interfaces';
+import { MatchDetails, OnExit, OutletContext, OnEnter, Params } from './interfaces';
 import { Router } from './Router';
+import { diffProperty } from '../widget-core/decorators/diffProperty';
+import { Handle } from '../core/Destroyable';
 
-export function isComponent<W extends WidgetBaseInterface>(value: any): value is Component<W> {
-	return Boolean(value && (typeof value === 'string' || typeof value === 'function' || typeof value === 'symbol'));
+export interface OutletProperties {
+	renderer: (matchDetails: MatchDetails) => DNode;
```
tomdye: Can this not return an array of `DNodes` ?

src/routing/interfaces.d.ts
agubler: Don't want to overload `properties`

src/routing/Outlet.ts
agubler: I think it should be able to.

PR #75, opened 2018-08-21
Widget class property `watch` decorator

### LINKED ISSUE #74: Add `watch` decorator that will invalidate when the property is set
**Enhancement**

Often widgets contain internal state that, when set, require the user to call `this.invalidate()` to ensure the widget is re-rendered.

Add a decorator for watching internal properties so that when they are set it will automatically call `invalidate`

```ts
class MyWidget extends WidgetBase {

    @watch()
    private _myState = 0;

    private _increment() {
         this._myState = this._myState + 1; // this automatically calls invalidate
    }

    protected render() {
        return v('div', [ 
            v('button', { onclick: this._increment }, [ 'Increment' ]),
            `${this._myState}`)
        ]);
    }
}
```

agubler:
**Type:** feature

**Description:**

Adds a watch decorator that will call `invalidate` when the property is set automatically.

Resolves #74

### DIFF
--- a/src/widget-core/README.md
+++ b/src/widget-core/README.md
@@ -125,6 +125,34 @@ class Hello extends WidgetBase<MyProperties> {
 
 New properties are compared with the previous properties to determine if a widget requires re-rendering. By default Dojo uses the `auto` diffing strategy, that performs a shallow comparison for objects and arrays, ignores functions (except classes that extend `WidgetBase`) and a reference comparison for all other values.
 
+#### Internal Widget State
+
+It is common for widgets to maintain internal state, that directly affects the results of the render output or passed as properties to child widgets. The most common pattern is that an action (often user initiated via an event) occurs which updates the internal state leaving the user to manually call `this.invalidate()` to trigger a re-render.
+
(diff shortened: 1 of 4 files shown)


src/widget-core/README.md
```diff
@@ -125,6 +125,34 @@ class Hello extends WidgetBase<MyProperties> {
 
 New properties are compared with the previous properties to determine if a widget requires re-rendering. By default Dojo uses the `auto` diffing strategy, that performs a shallow comparison for objects and arrays, ignores functions (except classes that extend `WidgetBase`) and a reference comparison for all other values.
 
+#### Internal Widget State
+
+It is common for widgets to maintain internal state, that directly affects the results of the render output or passed as properties to child widgets. The most common pattern is that an action (often user initiated via an event) occurs which updates the internal state leaving the user to manually call `this.invalidate()` to trigger a re-render.
+
+For class properties that always need to trigger a re-render when it's updated, a property decorator, `@watch` can be used, which implicitly calls `this.invalidate` each time the property is set.
```
mwistrand: "trigger a re-render when it's updated": "it's" => "they're"

PR #79, opened 2018-08-22
[WIP] Routing Prompt Widget

### LINKED ISSUE #73: Prompt widget for routing that renders a prompt before a route has transitioned
**Enhancement**

Add a new widget that can be used to render a view before a route transition has been completed based on a predicate. The render function would receive two functions that can be used to abort or continue the route transition.

For window reload/refresh, will need to use `onbeforeunload` and this only support the native browser confirm dialog.

agubler:
**Type:** feature

**Description:**

Provides a `Prompt` widget that can be used to prompt a user when they try to navigate away from a route and a the `shouldBlock` predicate returns `true`.

Resolves #73

### DIFF
--- a/src/routing/Prompt.ts
+++ b/src/routing/Prompt.ts
@@ -46,3 +46,5 @@ export class Prompt extends WidgetBase<PromptProperties> {
 		}
 	}
 }
+
+export default Prompt;
--- a/src/routing/Prompt.ts
+++ b/src/routing/Prompt.ts
@@ -1,0 +1,48 @@
+import global from '../shim/global';
+import WidgetBase from '../widget-core/WidgetBase';
+import { DNode } from '../widget-core/interfaces';
+import { Router } from './Router';
+import { Handle } from '../core/Destroyable';
+
+export interface PromptProperties {
+	shouldBlock: () => boolean;
+	renderer: (options: { continue: Function; block: Function }) => DNode | DNode[];
+	routerKey?: string;
+}
+
+export class Prompt extends WidgetBase<PromptProperties> {
+	private _handle: Handle;
+
+	private _onBeforeUnload = (event: Event) => {
+		if (this.properties.shouldBlock()) {
+			event.returnValue = true;
+		}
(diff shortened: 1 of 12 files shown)


src/routing/Prompt.ts
```diff
@@ -0,0 +1,50 @@
...
+			event.returnValue = true;
+		}
+	};
+
+	protected render(): DNode | DNode[] {
+		const item = this.registry.getInjector<Router>(this.properties.routerKey || 'router');
+		if (item) {
+			const router = item.injector();
+			if (!this._handle) {
+				this._handle = router.registerPrompt(this.properties.shouldBlock, () => {
```
JamesMilnerUK: Would this benefit from being inside it's own function i.e. `this._addHandle`?

src/routing/Prompt.ts
```diff
@@ -0,0 +1,50 @@
...
+
+	private _onBeforeUnload = (event: Event) => {
+		if (this.properties.shouldBlock()) {
+			event.returnValue = true;
+		}
+	};
+
+	protected render(): DNode | DNode[] {
+		const item = this.registry.getInjector<Router>(this.properties.routerKey || 'router');
+		if (item) {
```
JamesMilnerUK: Stylistic, but you could early return here to save an indent

src/routing/history/StateHistory.ts
```diff
@@ -41,11 +42,21 @@ export class StateHistory implements HistoryInterface {
...
+		return this._previous;
+	}
+
+	public replace(path: string) {
+		const value = ensureLeadingSlash(path);
+		this._window.history.replaceState({}, '', this.prefix(value));
+	}
+
 	private _onChange = () => {
 		const value = stripBase(this._base, this._window.location.pathname + this._window.location.search);
```
JamesMilnerUK: Worth putting pathname + search into a variable?

agubler:
Needs rebasing based on the changes to the router.

PR #86, opened 2018-08-29
Taking route parameters into account

### LINKED ISSUE #72: ActiveLink does not check route parameters
**Bug**

If you have two links that are identical except for the parameters, `ActiveLink` will mark them both as active.

Route Config:
```typescript
{
    path: 'my-route',
    outlet: 'my-route',
    children: [
        {
            path: '{suffix}',
            outlet: 'my-suffix'
        }
    ]
}
```

The links (using tsx),

```jsx
<ActiveLink to="my-route" params={ {"suffix": "test1"} } />
<ActiveLink to="my-route" params={ {"suffix": "test2"} } />
```

Both links will be marked as active when `my-route` is active.

Package Version: 3.0.0

**Expected behavior:**

The current params would be checked and only one link would be active.

**Actual behavior:**

All links to the outlet are active regardless of parameters.

rorticus:
**Type:** bug

**Description:**

Fix for `ActiveLink` not taking route parameters into account and falsely marking routes as active.

Resolves #72

### DIFF
--- a/src/routing/ActiveLink.ts
+++ b/src/routing/ActiveLink.ts
@@ -14,6 +14,10 @@ export interface ActiveLinkProperties extends LinkProperties {
 export class ActiveLink extends WidgetBase<ActiveLinkProperties> {
 	private _outletHandle: Handle | undefined;
 
+	private _paramsEqual(linkParams: any = {}, contextParams: any = {}) {
+		return Object.keys(linkParams).every((key) => linkParams[key] === contextParams[key]);
+	}
+
 	private _renderLink(isActive = false) {
 		let { activeClasses, classes = [], ...props } = this.properties;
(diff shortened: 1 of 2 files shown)


src/routing/ActiveLink.ts
```diff
@@ -14,6 +14,10 @@ export interface ActiveLinkProperties extends LinkProperties {
 export class ActiveLink extends WidgetBase<ActiveLinkProperties> {
 	private _outletHandle: Handle | undefined;
 
+	private _paramsEqual(linkParams: any = {}, contextParams: any = {}) {
```
agubler: This doesn't actually have to be on the `ActiveLink` instance

rorticus: Good point! I guess we can move this out into a fn.

PR #87, opened 2018-08-29
Ensure to match against the most specific route

### LINKED ISSUE #83: Order independent routing configuration
**Enhancement**

The order that routes are defined in the routing configuration currently affects which route will be matched when a path can match against multiple outlets, for example:

```ts
const config = [
    {
        path: '{param}',
        outlet: 'param'
    },
    {
        path: 'foo',
        outlet: 'foo'
    }
];
```

When the outlet `params` is defined first it will considered matched against the path `/foo` even though the `foo` outlet is a "more specific match".

The router should be able to determine which is the most specific outlet match, regardless of the configuration order.

agubler:
**Type:** feature/bug

**Description:**

For routing the first match route specified in the configuration will be considered "the one", even if there is a more exact match in a later route within the configuration.

Assuming a routing configuration:

```ts
const config = [
    {
        path: 'foo',
        outlet: 'foo-one'
    },
    {
        path: 'foo',
        outlet: 'foo-two',
        children: [
            {
                path: 'baz',
                outlet: 'baz'
            }
        ]
    },
    {
        path: '{bar}',
        outlet: 'param'
    },
    {
        path: 'bar',
        outlet: 'bar'
    }
];
```

### Current Behaviour

Routing to `/bar`, should match against the `bar` outlet as it's more specific (a string literal match) but currently would match against the `param` outlet because it is registered first.

Routing to `/foo/baz` would currently match against the outlet, `foo-one` with a type of partial, rather then correctly matching against the more exact outlet match `baz`.

### Updated Bahaviour

Routing to `/bar`, will match against the `bar` outlet as it's more specific (a string literal match).

Routing to `/foo/baz` will match against outlet `baz`.

Resolves #83

### DIFF
--- a/src/routing/Router.ts
+++ b/src/routing/Router.ts
@@ -25,6 +25,9 @@ export interface OutletEvent extends EventObject<string> {
 	action: 'enter' | 'exit';
 }
 
+const ROUTE_SEGMENT_SCORE = 7;
+const DYNAMIC_SEGMENT_PENALTY = 2;
+
 export class Router extends QueuingEvented<{ nav: NavEvent; outlet: OutletEvent }> implements RouterInterface {
 	private _routes: Route[] = [];
 	private _outletMap: { [index: string]: Route } = Object.create(null);
@@ -160,9 +163,9 @@ export class Router extends QueuingEvented<{ nav: NavEvent; outlet: OutletEvent 
 			}
 			for (let i = 0; i < segments.length; i++) {
 				const segment = segments[i];
-				route.score += 7;
+				route.score += ROUTE_SEGMENT_SCORE;
 				if (paramRegExp.test(segment)) {
-					route.score -= 2;
+					route.score -= DYNAMIC_SEGMENT_PENALTY;
 					route.params.push(segment.replace('{', '').replace('}', ''));
 					segments[i] = PARAM;
 				}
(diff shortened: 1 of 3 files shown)


src/routing/Router.ts
```diff
@@ -151,14 +152,17 @@ export class Router extends QueuingEvented<{ nav: NavEvent; outlet: OutletEvent
...
+				score: parentRoute ? parentRoute.score : 0,
 				onExit
 			};
 			if (defaultRoute) {
 				this._defaultOutlet = outlet;
 			}
 			for (let i = 0; i < segments.length; i++) {
 				const segment = segments[i];
-				if (typeof segment === 'string' && segment[0] === '{') {
+				route.score += 7;
```
rorticus: Might want to create constants for these numbers so we know what they mean!

src/routing/Router.ts
```diff
@@ -24,6 +25,9 @@ export interface OutletEvent extends EventObject<string> {
 	action: 'enter' | 'exit';
 }
 
+const ROUTE_SEGMENT_SCORE = 7;
```
JamesMilnerUK: These numbers look fairly arbitrary at first glance. Perhaps a comment to clarify?

agubler: They actually are fairly arbitrary, I went for a score that meant we could introduce other types of segments in the future (like maybe wild card, which would have an even smaller score).

PR #94, opened 2018-09-07
Add .finally() to Promise shim

### LINKED ISSUE #45: Add support in Promise for 'finally' method
@dasa commented on [Wed Jan 24 2018](https://github.com/dojo/shim/issues/129)

**Enhancement** 

Add support in Promise for `finally` method since it has now reached stage 4.

Related Issue: https://github.com/dojo/shim/issues/21#issuecomment-251464355

bryanforbes:
**Type:** feature

**Description:**

Adds `.finally()` method to Promise shim as defined in the [`Promise.prototype.finally` proposal](https://tc39.github.io/proposal-promise-finally/).

Resolves #45

### DIFF
--- a/src/shim/support/has.ts
+++ b/src/shim/support/has.ts
@@ -128,12 +128,11 @@ add(
 add('es-observable', () => typeof global.Observable !== 'undefined', true);
 
 /* Promise */
+add('es6-promise', () => typeof global.Promise !== 'undefined' && has('es6-symbol'), true);
+
 add(
-	'es6-promise',
-	() =>
-		typeof global.Promise !== 'undefined' &&
-		has('es6-symbol') &&
-		typeof global.Promise.prototype.finally !== 'undefined',
+	'es2018-promise-finally',
+	() => has('es6-promise') && typeof global.Promise.prototype.finally !== 'undefined',
 	true
 );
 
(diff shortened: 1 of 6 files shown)


src/shim/support/has.ts
```diff
@@ -128,7 +128,14 @@ add(
 add('es-observable', () => typeof global.Observable !== 'undefined', true);
 
 /* Promise */
-add('es6-promise', () => typeof global.Promise !== 'undefined' && has('es6-symbol'), true);
+add(
+	'es6-promise',
+	() =>
+		typeof global.Promise !== 'undefined' &&
+		has('es6-symbol') &&
+		typeof global.Promise.prototype.finally !== 'undefined',
```
matt-gadd: I don't think we should couple this to the `es6-promise` sniff, as for example are we not going to override the native on the global in say edge which has no `finally` support but does have `Promise`?

bryanforbes: The `finally` test should probably be its own test. However, if the promise shim has `finally` and the global one does not, `shim/Promise` either returns the global `Promise` or replaces the global one with our shimmed `Promise` and returns the shimmed one. If we return the global one in Edge, it won't have a `finally` method and will error out at runtime when a user tries to use `finally`. In this case, we have two options: return our shimmed `Promise` as we're doing now, or add `finally` to the global `Promise.prototype`. Which do you recommend?

PR #135, opened 2018-10-15
Update theme and animation classes to allow the output of this.theme

### LINKED ISSUE #134: extraClasses and animation classes typings do not match this.theme
Since `this.theme` can return null, using it to pass a themed class to `extraClasses` or `enterAnimation`/`exitAnimation` results in a type error. See this:
https://github.com/dojo/widgets/blob/master/src/grid/widgets/Cell.ts#L89
or this: https://github.com/dojo/widgets/pull/604/files#r225018265

smhigley:
**Type:** feature

**Description:**

Resolves #134 by using `SupportedClassName` for everything

### DIFF
--- a/src/widget-core/animations/cssTransitions.ts
+++ b/src/widget-core/animations/cssTransitions.ts
@@ -1,4 +1,4 @@
-import { VNodeProperties } from './../interfaces';
+import { VNodeProperties, SupportedClassName } from './../interfaces';
 
 let browserSpecificTransitionEndEventName = '';
 let browserSpecificAnimationEndEventName = '';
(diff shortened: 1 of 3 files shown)


agubler:
@smhigley Regarding the enter and exit animation, I believe that the vdom already guards against `falsy` values. This just needs the type being changed on the interface and couple of tests in vdom to prove it. Something like:

```ts
it('Does not invoke transition when null passed as enterAnimation', () => {
	const transition = {
		enter: stub(),
		exit: stub()
	};
	const [Widget, meta] = getWidget(v('div', [ v('span', { exitAnimation: null } )]));
	const r = renderer(() => w(Widget, {}));
	const div = document.createElement('div');
	r.mount({ domNode: div, sync: true, transition });
	meta.setRenderResult(v('div',[]));
	assert.isTrue(transition.exit.notCalled);
});
```

The `extraClasses` type change could do with some tests also to prove that it can handle `null` and `undefined` as needed.

agubler:
Superseded by #181 that adds tests to vdom and only adds the `SupportClassName` interface to the `exit/enter` animation properties.

`extraClasses` is going to be superseded by the new [classes API](#164 ) which already support `null` and `undefined` (leaving it will also encourage migration to `classes`)

PR #137, opened 2018-10-15
Emit the routing event after the routing has been completed

### LINKED ISSUE #136: Routing does not work when the renderer is in sync mode
**Bug**

When the projector is in sync mode then the components for outlets and links (anything related to routing) actually re-render before the routing navigation is actually completed. Meaning that an application is not correctly rendered.

agubler:
**Type:** bug

**Description:**

Ensure that the routing events are emitted after the routing navigation is completed.

Resolves #136

### DIFF
--- a/tests/routing/unit/ActiveLink.ts
+++ b/tests/routing/unit/ActiveLink.ts
@@ -293,17 +293,17 @@ describe('ActiveLink', () => {
 			router.setPath('/bar');
 			assert.strictEqual(
 				root.outerHTML,
-				'<div><div><a href="foo"></a><a class="bar" href="bar"></a><a href="baz"></a></div></div>'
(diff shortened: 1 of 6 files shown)


tests/routing/unit/ActiveLink.ts
```diff
@@ -244,4 +249,58 @@ describe('ActiveLink', () => {
 
 		assert.deepEqual(dNode2.properties.classes, []);
 	});
+
+	jsdomDescribe('integration tests', () => {
```
dylans: should this be `describe`, not `jsdomDescribe`? Perhaps I'm misunderstanding, but I thought `jsdomDescribe` was the type here?

agubler: Nope, this is the plugin to use jsdom.

PR #148, opened 2018-10-25
Fixes: Routing and Link Fixes

### LINKED ISSUE #145: Routing: Navigation to default route should be syncronous
**Bug**

Navigating to a default route hands off responsibility to the history adapters which use the appropriate mechanism, when using the `HashHistory` this uses the `hashchange` event which is asynchronous.

When calling the `router` instance's `setPath()` explicitly we should synchronously process the change and guard against repeated (surplus) processing in the adaptors.

### LINKED ISSUE #146: Routing: The router should generated correctly formatted links
**Bug**

`router.link()` generates links without the required prefix for the chosen adaptor. The correct prefix should be applied during the link generation.

### LINKED ISSUE #147: Routing: CTRL/CMD click should open the a `Link` in a new window
**Bug**

Currently links generated by the `Link` component do not support using `cmd` or `ctrl` (depending on the operating system) to open the link in a tab.

agubler:
**Type:** bug

**Description:**

A collection of fixes for the routing:

* Support ctrl/cmd click
* Ensure links are generated with appropriate prefixes
* Ensure routing is synchronous with `setPath` when using `HashHistory`
* Strip leading `#` character when using `StateHistory`

Resolves #147 
Resolves #146 
Resolves #145

### DIFF
--- a/src/routing/history/StateHistory.ts
+++ b/src/routing/history/StateHistory.ts
@@ -1,6 +1,20 @@
 import global from '../../shim/global';
 import { History as HistoryInterface, HistoryOptions, OnChangeFunction } from './../interfaces';
 
+const trailingSlash = new RegExp(/\/$/);
+
+function stripBase(base: string, path: string): string {
+	if (base === '/') {
+		return path;
+	}
+
+	if (path.indexOf(base) === 0) {
+		return path.slice(base.length - 1);
+	} else {
+		return '/';
+	}
+}
+
 export class StateHistory implements HistoryInterface {
 	private _current: string;
 	private _onChangeFunction: OnChangeFunction;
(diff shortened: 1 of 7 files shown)


src/routing/history/StateHistory.ts
```diff
@@ -14,26 +28,26 @@ export class StateHistory implements HistoryInterface {
...
-		if (baseEndsWithSlash && pathStartsWithSlash) {
-			return this._base + path.slice(1);
-		} else if (!baseEndsWithSlash && !pathStartsWithSlash) {
-			return `${this._base}/${path}`;
-		} else {
-			return this._base + path;
+		if (path[0] === '#') {
+			path = path.slice(1);
+		}
+		if (path[0] === '/') {
```
matt-gadd: this can be one if with the above right?

agubler: I did it this was to guard against `#/link` which needs to be `link`

PR #162, opened 2018-11-09
Exposing select in harness

rorticus:
**Type:** feature

**Description:**

We've already got the `select` utility, but this PR just exposes it to be a little more... friendly. Just adds a way to select DNodes directly from your harness. Once you have a DNode you can asset various things, call callbacks, etc.

```typescript
const h = harness(() => w(MyWidget, {
    key: "root",
    footer: v('button', { onclick: () => doSomething() }, 'footer')
}));
const [myWidget] = h.select("@root");

myWidget.properties.footer.onclick();
```

`select` will trigger a new render when you call it, so you can do things like,

```typescript
const foo = 'hello'
const h = harness(() => w(MyWidget, { foo }));
assert.isTrue(h.select('@span').length === 1);
foo = 'bar';
assert.isTrue(h.select('@span').length === 1);
```

### DIFF
--- a/src/testing/README.md
+++ b/src/testing/README.md
@@ -9,6 +9,7 @@ Simple API for testing and asserting Dojo widget's expected virtual DOM and beha
 -   [`harness.expect`](#harnessexpect)
 -   [`harness.expectPartial`](#harnessexpectpartial)
 -   [`harness.trigger`](#harnesstrigger)
+-   `harness.select`
 
 ## Features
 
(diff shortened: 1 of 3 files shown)


src/testing/README.md
```diff
@@ -230,3 +232,46 @@ const render = h.getRender();
...
+
+#### `harness.select`
+
+`harness.select` returns a list of DNodes, using the latest render, that match the provided selector. This can be useful if you need access to a widget's properties.
+
+```typescript
+select(selector: string);
+```
+
+-   `selector`: The selector to use while determining matching DNodes
```
agubler: mention it returns an array? and in the API above?

src/testing/README.md
```diff
@@ -9,6 +9,7 @@ Simple API for testing and asserting Dojo widget's expected virtual DOM and beha
 -   [`harness.expect`](#harnessexpect)
 -   [`harness.expectPartial`](#harnessexpectpartial)
 -   [`harness.trigger`](#harnesstrigger)
+-   `harness.select`
```
agubler: Make this a link?
```
[`harness.select`](#harnessselect)
```

agubler:
@rorticus I couple of comments on the readme, otherwise :+1:

agubler:
@rorticus there are some very tiny changes to the readme, if we can fix that up we can land this.

src/testing/README.md
rorticus: Done!

src/testing/README.md
rorticus: Updated the select api to be `select(selector: string): DNode[]`

PR #168, opened 2018-11-11
Process Exit Animations before removing parent nodes

### LINKED ISSUE #167: Exit Animations are missed when a parent node is removed
**Bug**

If a VNode has been created with an exit animation, when it is removed from the DOM the vdom renderer should execute the exit animation before removing the VNode or any parent from the DOM.

Assuming the example below, if the wrapper node is removed normally `removeChild` would be called on the `root` domNode. However if there is an `exitAnimation` in the widget tree, we cannot orphan the tree until the exit animation has been executed. 

```ts
render() {
    return v('div', { key: 'root' }, [
        v('div', { key: 'wrapper' }, [
            v('span', { exitAnimation: 'first' }
        ])
    ]);
}
```

We currently know if there is an exit animation nested in the widget tree, so we need to prevent orphaning the dom element and ensure that the exit animation is called before any parent nodes are removed.

agubler:
**Type:** bug

**Description:**

Nodes with an exit animation need to be processed before they are orphaned from the DOM. This change detects when there are child nodes with an exit animation and ensures that they are processed before any parent nodes are removed from the DOM.

Resolves #167

### DIFF
--- a/src/widget-core/animations/cssTransitions.ts
+++ b/src/widget-core/animations/cssTransitions.ts
@@ -26,12 +26,12 @@ function runAndCleanUp(element: HTMLElement, startAnimation: () => void, finishA
 
 	let finished = false;
 
-	let transitionEnd = function() {
+	let transitionEnd = function(event: Event) {
+		event.stopPropagation();
 		if (!finished) {
 			finished = true;
 			element.removeEventListener(browserSpecificTransitionEndEventName, transitionEnd);
 			element.removeEventListener(browserSpecificAnimationEndEventName, transitionEnd);
-
 			finishAnimation();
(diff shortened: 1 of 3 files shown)


agubler:
The more I think about this more I don't think this is the correct behaviour. For an exit animation to execute the user will need to ensure that the node is explicitly removed, rather than being removed implicitly by being orphaned higher in the widget tree.

An example is working with an application that has multiple "scenes/routes", the user navigates to a route that contains a deeply nested widget that has exit animations... Navigating to a new route shouldn't trigger the nested animation before the transition, the animations should only execute when dealing directly with the nodes in question with the route.

PR #169, opened 2018-11-11
Run onDetach for widgets synchronously

### LINKED ISSUE #165: Detach widgets synchronously
**Enhancement**

The current rendering logic assumes once a parent `VNode` has had it's `domNode` removed from it's parent then all subsequent children can be deferred until the client is idle using `requestIdleCallback`. 

The current exception is if a child has registered an `exitAnimation` which needs to be processed specifically to ensure the animation runs as expected.

However if the child tree contains a widget node then deferring the removal can be problematic, specifically when a meta might have been used. For example the `Resize` meta, as once a parent node has been detached from the dom tree it will cause the `Resize` meta to trigger a re-render for a widget that has actually been detached. Because the render is on the `requestAnimationFrame` it will be processed before the deferred cleanup runs on the `requestIdleCallback` causing unnecessary component renders and dom operations.

Additionally when an application is using containers to invalidate and inject state directly into sub trees of the application it can cause the same nodes to be processed by both the standard processing and also by the deferred clean up.

We should add an additional exception to only defer the clean up if there are no widget nodes in the sub tree that is being processed.

agubler:
**Type:** feature

**Description:**

Ensure that when widgets are removed from virtual dom tree that `onDetach` is called synchronously, only deferring cleanup to `requestIdleCallback` for purely VNode subtrees. VNode subtrees can be deferred because they will have already been orphaned,

Resolves #165

### DIFF
--- a/src/widget-core/vdom.ts
+++ b/src/widget-core/vdom.ts
@@ -25,6 +25,7 @@ export interface BaseNodeWrapper {
 	hasParentWNode?: boolean;
 	namespace?: string;
 	hasAnimations?: boolean;
+	hasWidgetChild?: boolean;
 }
 
 export interface WNodeWrapper extends BaseNodeWrapper {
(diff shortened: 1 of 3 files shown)


agubler:
Closing in favour of #271

PR #173, opened 2018-11-15
Add before and after hooks to store middleware

### LINKED ISSUE #90: Improve middleware for processes to allow universal functionality before executing commands
**Enhancement** 

It would be good to have somewhere to define functionality like handling expired sessions before running processes. Consider an application that hits the server in multiple processes. You need to be logged in to access API endpoints within the processes but your session could have expired. As it stands you need to handle this by running a shared command in every process that will do the session checking.  Having a universal way of handling this scenario via middleware would be slicker so that we can execute functionality before running the commands in a process.

Package Version: 3.0.1

maier49:
**Type:**feature

**Description:**
Adds the concept of "initializers" that can perform sync/async setup before a process is run. 
Resolves #90

### DIFF
--- a/src/core/README.md
+++ b/src/core/README.md
@@ -4,10 +4,10 @@ Provides a small set of base classes and core helpers for use within dojo/framew
 
 ## Features
 
-- [Destroyable](#destroyable)
-- [Evented](#evented)
-- [QueuingEvented](#queuingevented)
-- [util](#util)
+-   [Destroyable](#destroyable)
+-   [Evented](#evented)
+-   [QueuingEvented](#queuingevented)
+-   [util](#util)
 
 ### Destroyable
 
--- a/src/stores/README.md
+++ b/src/stores/README.md
@@ -28,7 +28,10 @@ An application store for dojo.
     -   [Optimistic Update Pattern](#optimistic-update-pattern)
     -   [Executing Concurrent Commands](#executing-concurrent-commands)
 -   [Middleware](#middleware)
+    -   [After Middleware](#after-middleware)
+    -   [Before Middleware](#before-middleware)
     -   [Applying Middleware to Multiple Processes](#applying-middleware-to-multiple-processes)
+    -   [Combining Before and After Callbacks](#combining-before-and-after-callbacks)
     -   [Local Storage Middleware](#local-storage-middleware)
 
 ---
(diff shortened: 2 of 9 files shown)


agubler:
Some general thoughts:

* As this is going to be released into v5.0.0, we can take a look at the existing API. Do you think it would be possible for middleware to support both scenarios, over adding a new concept in initialisers?
* Would it make more sense to manage the chaining/composition of the middleware internally, so that consumers don't need to use the `create*Decorator` functions, perhaps just taking an array of middlewares?
* It would be great if initialisers weren't always treated as async as it means a process with sync commands won't run sync?
* Do you think that the `payload` is the only thing that an initialiser will ever need? What about generic scenarios that might require the initialiser to mark some state in the store? `isLoading` as an example.
* It looks like there's some duplication across `createInitializerDecorator` and `createCallbackDecorator` which can probably be normalised (Possibly not relevant if we change how the middleware is passed)

maier49:
> Some general thoughts:
> 
> * As this is going to be released into v5.0.0, we can take a look at the existing API. Do you think it would be possible for middleware to support both scenarios, over adding a new concept in initialisers?

Would something like providing an object for middleware with 
```
{ 'before': {one or more before middlewares}, 'after': {one or more after middlewares} }
```
Make sense to you? I tried for an API where each middleware was run before and after, by receiving the payload and passing it to a `next` that returns the result. But it ended up feeling much more complicated to use when mostly you'd want one or the other. It also makes the order that they execute in less obvious.
> * Would it make more sense to manage the chaining/composition of the middleware internally, so that consumers don't need to use the `create*Decorator` functions, perhaps just taking an array of middlewares?

matt-gadd:
could we have the middleware itself return the `before` and `after`? that way users just use middleware and don't need to know where it has to go? it also means middleware can have a `before` and `after` without the user having to specify the middleware twice?

> @agubler
> 
> > Some general thoughts:
> > 
> > * As this is going to be released into v5.0.0, we can take a look at the existing API. Do you think it would be possible for middleware to support both scenarios, over adding a new concept in initialisers?
> 
> Would something like providing an object for middleware with
> 
> ```
> { 'before': {one or more before middlewares}, 'after': {one or more after middlewares} }
> ```
> Make sense to you? I tried for an API where each middleware was run before and after, by receiving the payload and passing it to a `next` that returns the result. But it ended up feeling much more complicated to use when mostly you'd want one or the other. It also makes the order that they execute in less obvious.
> 
> > 
could we have the middleware itself return the `before` and `after`? that way users just use middleware and don't need to know where it has to go. It also means middleware can have a `before` and `after` without the user having to specify the middleware twice.

matt-gadd:
@maier49 apologies for the late review on this. this looks really good to me, I just have one question on the ordering of the `after` calls. Do we think it should be in reverse order like in the PR and why? In my head I can't quite decide what the correct ordering should be, and which is less surprising to the end user.

maier49:
@matt-gadd The thought process for the order is borrowed from some request hooks as the last interceptor that modifies the request might need to be the first to handle the response, for example for encrypting/decrypting. However, that doesn't really apply the same way to store processes, so I could be convinced either way on the ordering.

agubler:
@maier49 I leaning towards the `before` and `after` middlewares being executed in the order that they are defined. If I had a process that defined middleware that only executed after the process, i think it would be surprising if they executed in reverse order.

```ts
const myProcess = createProcess('example', [ command ], [ afterOne, afterTwo ]);
```

maier49:
@agubler That's updated now

PR #180, opened 2018-11-22
Update routing README

### LINKED ISSUE #143: Documentation out of date
I get the following error configuring  the route.
routing

I followed the example  in the documentation: https://dojo.io/docs/index.html#doc--dojo__framework__v4_0_0__src__routing__README_md

The following approach works: https://github.com/dojo/examples/blob/master/realworld/src/main.ts

DoJo Version: 4.0

agubler:
**Type:** bug

**Description:**

Corrects the routing README

Resolves #143

### DIFF
--- a/src/routing/README.md
+++ b/src/routing/README.md
@@ -124,7 +124,7 @@ import { MemoryHistory } from '@dojo/framework/routing/MemoryHistory';
 const router = new Router(config, MemoryHistory);
 ```
 
-Once the router has been created with the application route configuration, it needs to be made available to all the components within your application. This is done using a `Registry` from `@dojo/framework/widget-core/Registry` and defining an `Injector` that contains the `router` instance as the `payload`. This `Injector` is defined using a known key, by default the key is `router` but this can be overridden if desired.
+Once the router has been created with the application route configuration, it needs to be made available to all the components within your application. This is done using a `Registry` from `@dojo/framework/widget-core/Registry` and defining an `Injector` that contains returns the `router` instance. This `Injector` is defined using a known key, by default the key is `router` but this can be overridden if desired.
 
 ```ts
 import { Registry } from '@dojo/framework/widget-core/Registry';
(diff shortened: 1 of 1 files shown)


src/routing/README.md
```diff
@@ -124,7 +124,7 @@ import { MemoryHistory } from '@dojo/framework/routing/MemoryHistory';
 const router = new Router(config, MemoryHistory);
 ```
 
-Once the router has been created with the application route configuration, it needs to be made available to all the components within your application. This is done using a `Registry` from `@dojo/framework/widget-core/Registry` and defining an `Injector` that contains the `router` instance as the `payload`. This `Injector` is defined using a known key, by default the key is `router` but this can be overridden if desired.
+Once the router has been created with the application route configuration, it needs to be made available to all the components within your application. This is done using a `Registry` from `@dojo/framework/widget-core/Registry` and defining an `Injector` that contains returns the `router` instance. This `Injector` is defined using a known key, by default the key is `router` but this can be overridden if desired.
```
dylans: "defining an `Injector` that contains returns the `router` instance" I could read this a few different ways, so I'd suggest adding a word or two for clarity

src/routing/README.md
```diff
@@ -133,14 +133,19 @@ import { Injector } from '@dojo/framework/widget-core/Injector';
...
+registry.defineInjector('router', () => {
+	router.on('nav', () => invalidator());
+	return () => router;
+};
 ```
 
-Finally, the `registry` needs to be made available to all widgets within the application by setting it as a `property` to the application's top-level `Projector` instance.
+**Note:** Routing provides a [convenience method for registering the router](#router-context-injection).
+
+Finally, the `registry` needs to be made available to all widgets within the application by passing it to the `.mount()` method of the vdom `renderer`
```
dylans:
```suggestion
Finally, the `registry` needs to be made available to all widgets within the application by passing it to the `.mount()` method of the vdom `renderer`.
```

src/routing/README.md
agubler: Updated

PR #213, opened 2019-01-04
Provide a state object in the command request that can be directly modified to update store state

### LINKED ISSUE #40: Proxy adapter instead of having to manually create operations
@matt-gadd commented on [Tue Jun 26 2018](https://github.com/dojo/stores/issues/177)

**Enhancement**
If we leverage Proxies, I think we can just about track the appropriate paths and changes to allow the user to simply do object and array accessing and setting rather than having to use the `get` helper and returning the associated operations.

The following would go from:
```ts
function calculateCountsCommand({ get, path }: CommandRequest) {
	const todos = get(path('todos'));
	const completedTodos = todos.filter((todo: any) => todo.completed);
	const operations = [
		replace(path('activeCount'), todos.length - completedTodos.length),
		replace(path('completedCount'), completedTodos.length)
	];
	return operations;
}
````
to:
```ts
function calculateCountsCommand(state) {
	const completedTodos = state.todos.filter((todo: any) => todo.completed);
	state.activeCount = state.todos.length - completedTodos.length;
	state.completedCount = completedTodos.length;
}
```

This would work via the `state` object being passed in being Proxy wrapped and everything off it. Every time a set is called we'd create the appropriate operation.

A few caveats:
* Won't work in IE11 (so we need to support the existing way)
* The functions no longer take an input and return an output (would have to think about the testing implications)
* Performance?

maier49:
**Type:** feature

**Description:**

This leverages proxies (where available), to add a `state` property to the command request object which can be modified directly to create operations that will update the store. Of the concerns raised on the original issue performance seems to be the biggest one, at least using this implementation. A simple operation that modifies one property two levels deep in an object takes about twice as long due to the time to create the proxy, regardless of whether it's used. Using the proxy slows down the operation further. 

Resolves #40


src/stores/process.ts
```diff
@@ -29,7 +32,7 @@ export interface CommandFactory<T = any, P extends object = DefaultPayload> {
  * Command that returns patch operations based on the command request
  */
 export interface Command<T = any, P extends object = DefaultPayload> {
-	(request: CommandRequest<T, P>): Promise<PatchOperation<T>[]> | PatchOperation<T>[];
+	(request: CommandRequest<T, P>): Promise<PatchOperation<T>[]> | PatchOperation<T>[] | void;
```
agubler: I think this needs to be `void | Promise<void>` for async commands when using the `state` object.

maier49:
I updated this to use a getter for state and throw an exception if used in IE. It seemed preferable to an alternative factory to me for a couple reasons:

* The user will still be informed that this is not usable in IE as soon as they run the code
* There's no risk of accidentally using the wrong factory in other browsers and not realizing the `state` object is available
* It allows the `apply` API to be consistent.

Because of the way the patches are applied providing a `state` object in IE  and performing a diff on it would be extremely slow, as we'd have to recursively build and diff it, and then still create the patch operations to do all those changes all over again. It would also require changing or violating the store API as we'd need root access to the state object.

agubler:
Nice work @maier49

src/stores/process.ts
agubler: @maier49 Did you see this comment?

maier49: @agubler Could have sworn I did that, sorry. It's done now!

PR #214, opened 2019-01-06
Update createCallbackDecorator to convert old middleware API to new API

agubler:
**Type:** feature

**Description:**

When creating stores middleware, ideally consumers would use the `createCallbackDecorator` factory, with the breaking API change to support `before` and `after` middlewares we can leverage this factory to convert the old API to the new API enabling a seem less upgrade.

I wasn't sure if about adding a warn about using the old API, at the moment this is included behind the `dojo-debug` flag which will only execute in development builds. Open to suggestions 😄

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -96,9 +96,9 @@
       }
     },
     "@theintern/leadfoot": {
-      "version": "2.2.3",
-      "resolved": "https://registry.npmjs.org/@theintern/leadfoot/-/leadfoot-2.2.3.tgz",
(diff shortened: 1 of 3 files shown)


agubler:
Note: need to update the warn to the real migration URL when it exists.

agubler:
@maier49 I've updated the `createCallbackDecorator` which was used before to create composable middlewares to convert middlewares to the new API seemlessly. Do you mind taking a look?

Issue #237, opened 2019-02-01
The ResizeObserver shim should not add to `global`


dasa:
**Bug ** 

Package Version: 5.0

**Code**

https://github.com/dojo/framework/blob/master/src/shim/ResizeObserver.ts#L34-L38

**Expected behavior:**

It returns the ponyfill: https://github.com/que-etc/resize-observer-polyfill#usage-example

**Actual behavior:**

It sets the ponyfill on `global` and then returns that. This also doesn't work if using an AMD loader that uses the non-ES6 module since `default` is then `undefined`.

matt-gadd:
The expected behaviour is to be on the global for shim and will be going forward as there are not ponyfills for everything out there externally. See `Pointer Events`, `IntersectionObserver`, `fetch`, `Promise`, `Symbol`, `WebAnimation` and maybe more that now align the same behaviour. Being on the global also allows us to completely elide the entire module during the build. 

It looks like we need to update the readme to clarify this, as it currently contradicts the above statement (other than for `Promise` and `Symbol`). We will likely be renaming `shim` to `platform` in 6.0.0 also.

We would definitely love to see a fix for the AMD bug though if that is an issue though.

dasa:
> We would definitely love to see a fix for the AMD bug though if that is an issue.

OK, I can try to create a PR for this. The easy fix would be to enable `esModuleInterop` in the `tsconfig`, but that requires TS 2.7. Would you prefer a check inside the shim itself?

matt-gadd:
@dasa we've been keen to try and support as far back as our initial TS version while we reasonably can. Perhaps switching from:
```ts
import Resize from 'resize-observer-polyfill';
```
to
```ts
import * as Resize from 'resize-observer-polyfill';
```
in the shim will work for now?

dasa:
> The expected behaviour is to be on the global for shim

Does this include array, string, Map, etc?

matt-gadd:
At the moment no, we only aligned the newest external polyfills/ponyfills to be consistent (`Promise` and `Symbol` were the only existing ones that were on the global). 

The array and string implementations are more problematic because they do not have the same api shape due to not being on the primitives prototype. 

We do have plans in general to try and make this more consistent on the road to v6. We could do with being more thorough on stating the end goal for shim/platform in documentation so apologies on the darkness there!

dasa:
> Would you prefer a check inside the shim itself?

PR created

PR #238, opened 2019-02-04
Add support for using resize-observer-polyfill's UMD module

dasa:
**Type:** bug

**Description:**

The shim returns `undefined` if using an AMD loader that uses the polyfill's UMD module since `default` is then `undefined`.

Resolves https://github.com/dojo/framework/issues/237#issuecomment-459818424

### DIFF
--- a/src/shim/ResizeObserver.ts
+++ b/src/shim/ResizeObserver.ts
@@ -1,7 +1,7 @@
 import global from './global';
 import has from '../has/has';
 `!has('build-elide')`;
-import Resize from 'resize-observer-polyfill';
+import * as Resize from 'resize-observer-polyfill';
 
 export interface DOMRectReadOnly {
 	readonly x: number;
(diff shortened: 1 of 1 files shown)


src/shim/ResizeObserver.ts
```diff
@@ -33,7 +33,8 @@ export interface ResizeObserver {
 
 if (!has('build-elide')) {
 	if (!global.ResizeObserver) {
-		global.ResizeObserver = Resize;
+		// default is undefined when UMD module is used
+		global.ResizeObserver = Resize.default || Resize;
```
matt-gadd: I don't think you need this change, when doing `* as Resize` it works correctly in non-umd environments.

dasa: Isn't `default` the same as a named export when doing a namespace import?

dasa: I get this error when I test this with a Dojo Webpack build without changing this line:

`TypeError: Object is not a constructor (evaluating 'new a.default')`

Also `window.ResizeObserver` is set to an Object that has a `default` property.

matt-gadd: @dasa yup i was being silly and testing this in chrome which has native support 😂 . thanks!

matt-gadd:
@dasa sorry it took so long to land. we'll be cutting a minor release this week for 5, thanks for fixing!

PR #240, opened 2019-02-05
Make sure we support assertion key for tsx in assertion templates

matt-gadd:
**Type:** bug

**Description:**
the tilde `~` is not supported as an attribute in `tsx`, so can't use the assertion key. This adds an also support attribute name of `assertion-key` so it can be used in tsx. note: you still select using the tilde `~`


matt-gadd:
@agubler done, thanks! also fixed the outdated readme for the last pr.

Issue #48, opened 2018-07-30
Potential improvements to state management performance, possibly making it pluggable


agubler:
@maier49 commented on [Fri May 11 2018](https://github.com/dojo/stores/issues/173)

For large state stores, things like ImmutableJS are recommended by the redux team as a way to improve update performance. On the other hand ImmutableJS can be quite slow for operations like finding a value. 

Making the actual state management portion of the store pluggable would allow a user to implement something that suits their needs without making additional assumptions.

Alternatively, or in addition to this, we could look into general performance improvements such as doing updates in a web worker, since we completely control the "reduction" part of the store life cycle.

maier49:
I've got a WIP branch where I've ripped out "State" as a pluggable interface in the store, and an ImmutableJS implementation. It's at the point of compiling but not running yet.

That's here: https://github.com/maier49/framework/tree/48-patch-performance

JamesMilnerUK:
### Some thoughts about putting Stores in a worker

I've been exploring how to put the Store in a worker. @agubler previous attempted this on a branch https://github.com/agubler/stores/tree/store-worker-poc. 

I had my own attempt and so far the problems I hit trying to achieve this were:

- Stores is currently a sync API, using a worker forces this to be Async for `get` and `apply` as a minimum. 
- `framework` currently compiles to UMD, but has no support for global fallbacks (see https://github.com/Microsoft/TypeScript/issues/8436). This means if your code uses `import` or `export` the whole file will be wrapped in a UMD wrapper which in turn means that when you point the worker at the file it never executes. If we want to share code we'd leverage Webpack/Rollup to build a worker specific file with all the dependencies inlined in a bundle.
- Since we can't reuse code (at least not easily) this leads to lots of duplication and potentially inconsistencies in interfaces. Long term this might actually be quite problematic if they drifted too much.
- I'm not sure what the implications of using a Web Worker having testing with Intern. Might be fine with browser based testing but Node would struggle unless without some mocking I would assume.

My attempts can be seen here:

https://github.com/JamesMilnerUK/framework/tree/worker

[...]

PR #242, opened 2019-02-05
Factor out a `MutableState` interface from the store and provide an ImmutableJS implementation

maier49:
**Type:** feature

**Description:**

Related to #48 

This creates a `MutableState` interface that has the properties of `State` as well as the `apply` function. It then breaks out the parts of store that implemented the `MutableState` interface into a default implementation, and allows an alternative implementation to be passed in to the constructor.

Includes an implementation using ImmutableJS. So far testing the performance impact of using ImmutableJS has not shown significant differences but I have not thoroughly tested it.

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,6 +48,7 @@
     "css-select-umd": "1.3.0-rc0",
     "diff": "3.5.0",
     "globalize": "1.4.0",
+    "immutable": "3.8.2",
     "intersection-observer": "0.4.2",
     "pepjs": "0.4.2",
     "resize-observer-polyfill": "1.5.0",
@@ -66,7 +67,6 @@
     "@types/selenium-webdriver": "^3.0.8",
     "@types/sinon": "~4.1.2",
     "benchmark": "^1.0.0",
-    "immutable": "3.8.2",
     "bootstrap": "^3.3.7",
     "chromedriver": "2.40.0",
     "cldr-data": "32.0.1",
@@ -83,9 +83,6 @@
     "selenium-webdriver": "3.6.0",
     "sinon": "~4.1.3",
     "typescript": "~2.6.2"
-  },
-  "optionalDependencies": {
-    "immutable": "3.8.2"
   },
   "lint-staged": {
     "*.{ts,tsx}": [
--- a/package.json
+++ b/package.json
@@ -48,7 +48,7 @@
     "css-select-umd": "1.3.0-rc0",
     "diff": "3.5.0",
     "globalize": "1.4.0",
-    "immutable": "^4.0.0-rc.12",
+    "immutable": "3.8.2",
     "intersection-observer": "0.4.2",
     "pepjs": "0.4.2",
     "resize-observer-polyfill": "1.5.0",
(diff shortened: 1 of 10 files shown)


maier49:
Looks like this version of immutable has different behavior in IE than other browsers so I'll have to sort that out

maier49:
Switched to using the latest official version of Immutable which doesn't suffer from the same issue.

package.json
```diff
@@ -52,6 +52,7 @@
     "css-select-umd": "1.3.0-rc0",
     "diff": "3.5.0",
     "globalize": "1.4.0",
+    "immutable": "3.8.2",
```
agubler: Should we include immutable as a dependency, or make it a dev dependency and also an optional dependency?

maier49: I made it an optional and dev dependency. In the code I just wrapped the require in a `try`. I'm assuming that we expect that module will fail if Immutable isn't installed, but I'm not sure how we want to handle that. I didn't see anywhere else we're using optional dependencies so let me know if that's the wrong approach.

src/stores/README.md
```diff
@@ -567,6 +568,31 @@ In this example, `commandOne` is executed, then both `concurrentCommandOne` and
...
+The default state implementation is reasonably optimized and in most circumstances will be sufficient.
+If a particular use case merits an alternative implementation it can be provided to the store constructor
+
+```ts
+const store = new Store({ state: myStateImpl });
+```
+
+#### ImmutableState
+
+An implementation of the `MutableState` interface that leverages [Immutable](https://github.com/immutable-js/immutable-js) under the hood is provided as
```
agubler: Is this formatting a little off with line endings?

package.json
agubler: Perhaps we should put something in the README to say that they would need to install immutable.... Or maybe having it as a dev isn't the end of the world... it's not going to affect anything unless they use it anyway.

tests/stores/unit/middleware/localStorage.ts
```diff
@@ -62,12 +62,12 @@ describe('middleware - local storage', (suite) => {
 	it('should load from local storage', () => {
 		global.localStorage.setItem(LOCAL_STORAGE_TEST_ID, '[{"meta":{"path":"/counter"},"state":1}]');
 		load(LOCAL_STORAGE_TEST_ID, store);
-		assert.deepEqual((store as any)._state, { counter: 1 });
+		assert.deepEqual((store as any)._state._state, { counter: 1 });
```
agubler: A side effect of the impl is having state stored in a nested `_state._state` object. Do you think that it should probably be something like `_adapter._state`?

package.json
maier49: Your call. Since it should be left out if they're not using it I'd be fine with it being a dependency.

agubler: yeah let's put it in the deps again. sorry!

maier49: No problem. Reverted

src/stores/README.md
maier49: I moved it to one line to avoid any issues.

tests/stores/unit/middleware/localStorage.ts
maier49: `_adapter` seems reasonable to me. I've updated it

PR #243, opened 2019-02-08
Adding stackable customElement decorators

rorticus:
**Type:** bug / feature

**Description:**

In preparation for https://github.com/dojo/webpack-contrib/issues/43 we need to make the `customElement` decorator stackable.  By default, everything that inherits `WidgetBase` will get a `@customElement` decorator with sensible defaults. So,

```ts
export interface MyProps {
    prop: string;
}

class WorstWidgetEver extends WidgetBase<MyProps> { }
```

would become,

```ts
export interface MyProps {
    prop: string;
}

@customElement({
   tag: 'worst-widget-ever',
   attributes: ['prop']
})
class WorstWidgetEver extends WidgetBase<MyProps> { }
```

To customize the declaration, you could add your own custom element decorator, which would look like this,

```ts
export interface MyProps {
    prop: string;
}

@customElement({ tag: 'best-widget-ever' })
class WorstWidgetEver extends WidgetBase<MyProps> { }
```

and after transformation would look like this,

```ts
export interface MyProps {
    prop: string;
}

@customElement({ tag: 'best-widget-ever' })
@customElement({
   tag: 'worst-widget-ever',
   attributes: ['prop']
})
class WorstWidgetEver extends WidgetBase<MyProps> { }
```

This PR specifically changes the `customElement` decorator to not clobber any previously set values, but instead merge them in.  

Additionally, this PR also moves the default values out of the `customElement` decorator and into the `registerCustomElement` function. When `cli-build-widget` adds the decorator manually, it doesn't actually add the decorator, but instead adds the property to the prototype chain directly. This is done to avoid adding imports to the file, but has a downside of not being able to apply the same defaults that the decorator can. The fix of course is just to apply the defaults somewhere else (where the values are _used_ rather than where they are defined)

Doesn't resolve, but is related to https://github.com/dojo/webpack-contrib/issues/43

### DIFF
--- a/src/widget-core/decorators/customElement.ts
+++ b/src/widget-core/decorators/customElement.ts
@@ -38,13 +38,6 @@ export interface CustomElementConfig<P extends object = { [index: string]: any }
  * registers that custom element.
  */
 export function customElement<P extends object = { [index: string]: any }>(config: CustomElementConfig<P>) {
-	const defaults = {
-		properties: [],
-		attributes: [],
-		events: [],
-		childType: CustomElementChildType.DOJO,
-		registryFactory: () => new Registry()
-	};
 	// rename "tag" to "tagName"
 	const { tag: tagName, ...configRest } = config;
 	const userDefinedConfig: CustomElementConfig & { tagName?: string } = configRest;
@@ -55,7 +48,6 @@ export function customElement<P extends object = { [index: string]: any }>(confi
 
 	return function<T extends Constructor<any>>(target: T) {
 		target.prototype.__customElementDescriptor = {
-			...defaults,
 			...target.prototype.__customElementDescriptor,
 			...userDefinedConfig
 		};
--- a/src/widget-core/decorators/customElement.ts
+++ b/src/widget-core/decorators/customElement.ts
@@ -45,14 +45,18 @@ export function customElement<P extends object = { [index: string]: any }>(confi
 		childType: CustomElementChildType.DOJO,
 		registryFactory: () => new Registry()
 	};
+	// rename "tag" to "tagName"
 	const { tag: tagName, ...configRest } = config;
-	const updatedConfig = { ...(tagName ? { tagName } : {}), ...configRest };
+	const userDefinedConfig = {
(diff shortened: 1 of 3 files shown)


src/widget-core/decorators/customElement.ts
```diff
@@ -37,22 +37,22 @@ export interface CustomElementConfig<P extends object = { [index: string]: any }
...
+export function customElement<P extends object = { [index: string]: any }>(config: CustomElementConfig<P>) {
+	const defaults = {
+		properties: [],
+		attributes: [],
+		events: [],
+		childType: CustomElementChildType.DOJO,
+		registryFactory: () => new Registry()
+	};
+	const { tag: tagName, ...configRest } = config;
+	const updatedConfig = { ...(tagName ? { tagName } : {}), ...configRest };
```
agubler: I find this really hard to follow what exactly is going on here... Perhaps it could be more verbose but end up clearer?

src/widget-core/decorators/customElement.ts
```diff
@@ -37,22 +37,22 @@ export interface CustomElementConfig<P extends object = { [index: string]: any }
...
 	return function<T extends Constructor<any>>(target: T) {
 		target.prototype.__customElementDescriptor = {
-			tagName: tag,
-			attributes,
-			properties,
-			events,
-			childType,
-			registryFactory
+			...defaults,
+			...target.prototype.__customElementDescriptor,
```
agubler: Is this always an object at this point?

agubler: Ignore me, doesn't matter does it... spreading `{ ...undefined }` results in an empty object it seems.

src/widget-core/decorators/customElement.ts
rorticus: @agubler I made some subtle changes here and added a comment. I mostly just removed the ternary (your comment below about undefined means this should destructure fine as an empty object if tag name isn't specified).  Anyways, hopefully this should be a little clearer now.

PR #247, opened 2019-02-13
Add convenience methods for children setting in Assertion Templates

matt-gadd:
**Type:** feature

**Description:**
Adds convenience methods (insertBefore, insertAfter, prepend, append and replace) to the Assertion Template api. This should make the api a little more discoverable over the third argument on `setChildren` and `insertChildren`, but functionally no different.

I've also renamed `insertChildren` to `insertSiblings` as the former was incorrect technically.

### DIFF
--- a/src/testing/README.md
+++ b/src/testing/README.md
@@ -320,7 +320,12 @@ Here we're using the `setChildren()` api on the baseAssertion, and we're using t
 Assertion Template has the following api's:
 
 ```
+insertBefore(selector: string, children: DNode[]): AssertionTemplateResult;
+insertAfter(selector: string, children: DNode[]): AssertionTemplateResult;
 insertChildren(selector: string, children: DNode[], type?: 'before' | 'after'): AssertionTemplateResult;
+append(selector: string, children: DNode[]): AssertionTemplateResult;
+prepend(selector: string, children: DNode[]): AssertionTemplateResult;
+replace(selector: string, children: DNode[]): AssertionTemplateResult;
 setChildren(selector: string, children: DNode[], type?: 'prepend' | 'replace' | 'append'): AssertionTemplateResult;
 setProperty(selector: string, property: string, value: any): AssertionTemplateResult;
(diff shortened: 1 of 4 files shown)


src/testing/README.md
```diff
@@ -320,7 +320,12 @@ Here we're using the `setChildren()` api on the baseAssertion, and we're using t
 Assertion Template has the following api's:
 
 ```
+insertBefore(selector: string, children: DNode[]): AssertionTemplateResult;
```
agubler: With the convenience APIs, do you think that we should make the naming more specific to avoid any possible confusion on usage? i.e. `appendChildren`, `prependChildren`, `replaceChildren`, `insertChildrenAfter` and `insertChildrenBefore`?

matt-gadd: I think at that point you don't gain much over the `setChildren` and `insertChildren` api's. I figured it seemed obvious enough it was operating with nodes and I guess the signature of taking an array of `DNodes[]`, probably covers it? I'm happy to close this PR if we think thats not the case though?

matt-gadd: I probably need to adjust the `insertChildren` api too anyway, as it doesn't really insert children, it inserts siblings...

agubler: That's true re `insertChildren`. I see your point about the signature helping to avoid confusion for their purpose, it was just that all the existing APIs were explicit so was caught a bit off guard with additional APIs without the `property`/`children` suffix.

matt-gadd: I think the only reason the other api's were/are explicit is because `set` clearly does conflict across the 2, i'm not sure the ones i've added here really do (perhaps `replace` at worst does?).

agubler: But also true, at that point they don't really add much value over the existing APIs.

agubler: Let's get this in, I think you're right that the naming and then the signature should be clear enough about the APIs usage. I do think that they offer a nice convenience over using the more verbose `setChildren` and `insertSibling` API.

PR #264, opened 2019-02-20
add validation meta

nicknisi:
**Type:** feature

* [x] There is a related issue (dojo/widgets#664)

**Description:**

This PR adds a new meta that specifically returns information from the browser's form validation API. This can be used to target HTML form elements, such as `<input type="email">`, and will return an object containing information about whether the element passes validation and whether it contains a message.

#### Invalid Example

Given the following:

```tsx
render() {
    const value = 'test';
    const validity = this.meta(InputValidity).get('input', value);
    render (
        <input type="email" key="input" value={value}/>
    )
}
```
This meta will return the following:

```typescript
{
    valid: false,
    message: "Please include an '@' in the email address. 'test' is missing an '@'."
}
```

#### Valid Example

Given the following:

```tsx
render() {
    const value = 'test@example.com';
    const validity = this.meta(InputValidity).get('input', value);
    render (
        <input type="email" key="input" value={value}/>
    )
}
```
This meta will return the following:

```typescript
{
    valid: true,
    message: ""
}
```

### DIFF
--- a/src/widget-core/meta/InputValidity.ts
+++ b/src/widget-core/meta/InputValidity.ts
@@ -1,0 +1,22 @@
+import Base from './Base';
+
+export class InputValidity extends Base {
+	get(key: string | number, value: string) {
+		const node = this.getNode(key) as HTMLFormElement | undefined;
+
+		if (!node) {
+			return { valid: undefined, message: '' };
+		}
+
+		if (value !== node.value) {
+			setTimeout(() => this.invalidate());
+		}
+
+		return {
+			valid: node.validity.valid,
+			message: node.validationMessage
+		};
+	}
+}
+
+export default InputValidity;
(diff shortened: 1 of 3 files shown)


src/widget-core/meta/InputValidity.ts
```diff
@@ -0,0 +1,26 @@
...
+		if (!node) {
+			return { valid: undefined, message: '' };
+		}
+
+		if (value !== node.value) {
+			// if the vdom is out of sync with the real dom our
+			// validation check will be one render behind.
+			// Call invalidate on the next loop.
+			setTimeout(() => this.invalidate());
+			return { valid: undefined, message: '' };
```
tomdye: could this not cause a flickering?

tomdye: I don't think you need to return her at all to be honest Nick, you weren't doing so in the POC.

nicknisi: Oops, that should actually be removed.

PR #278, opened 2019-02-27
No longer throw by default on unregistered has() checks 

matt-gadd:
**Type:** feature

**Description:**
No longer throw when performing has() checks on unregistered features by default. Throwing by default is quite annoying, especially as we now leverage a lot of `has()` checks as part of the build tool where they are mostly progressive, so when using without the build tool (ie in a codesandbox) we'd have to write different code.

An additional strict param to enable the old functionality has been added.


src/shim/support/decorators.ts
```diff
@@ -10,6 +10,6 @@ import has from '../../has/has';
 export function hasClass(feature: string, trueClass: Function, falseClass: Function): ClassDecorator {
```
matt-gadd: i have absolutely no idea why this module exists and we should remove it in 6.

PR #286, opened 2019-03-05
Routing Reference Guide

agubler:
**Type:** feature

**Description:**

Work in progress for the routing reference guide.


docs/en/routing/basic-usage.md
```diff
@@ -0,0 +1,490 @@
...
+```ts
+export default [
+	{
+		path: 'home/{page}?{queryOne}&{queryOne}',
+		outlet: 'home'
+	}
+];
+```
+
+The query params are injected into matching `Outlet` `renderer` properties accessed from `queryParams` property.
```
sbinge: Might be worth illustrating use of `queryParams` in tandem with `params`.

Also there's a duplicate `queryOne` param in the above route - which raises another interesting question of how the framework handles such a definition. Is it an error? Or would it create an array if more than one value is given?

docs/en/routing/basic-usage.md
```diff
@@ -0,0 +1,490 @@
...
+
+## Default Route & Parameters
+
+To add a default route for your application update the routing configuration to include `defaultRoute: true` for the preferred route.
+
+>src/routes.ts
+```ts
+export default [
+	{
+		path: 'home',
```
sbinge: What's the significance of a specific path given for a default route? A few more words explaining how a default route differs from regular routes might be useful (triggers on '/' and '/home'? or triggers on '/home' and any other route not explicitly given in `routes.ts`?).

docs/en/routing/basic-usage.md
agubler: That is a typo on my part, but it's a good point - currently the router doesn't have multiple query params but I can't think why we could support arrays for multiple query params.

docs/en/routing/basic-usage.md
```diff
@@ -0,0 +1,490 @@
+# Basic Usage
+
+## Adding Routing to an Application
+
+Begin by adding an initial route configuration, that defines a single route that maps to an identifier known as an `outlet`.
```
matt-gadd: I know this is structured similar in order to the I18n one. but I think it'd be better to start with registering the router first in this scenario. rather than talking about/referencing an outlet immediately

docs/en/routing/basic-usage.md
```diff
@@ -0,0 +1,490 @@
...
+
+const registry = new Registry();
+// creates a router with the routes and registers the router with the registry
+registerRouterInjector(routes, registry);
+
+const r = renderer(() => <App />);
+r.mount({ registry });
+```
+
+Use the `Outlet` widget to control what widgets to render when the path for an outlet id is matched, the example will render `<div>Home</div>` for the `home` outlet.
```
matt-gadd: Give context where we are here? something like:
> To display something when a route is matched, we use a mechanism called an Outlet. So lets add one to show `Home` when the `home` route is visited

docs/en/routing/basic-usage.md
```diff
@@ -0,0 +1,490 @@
...
+export default class App extends WidgetBase {
+	protected render() {
+		return (
+			<div>
+				<Outlet id="home" renderer={() => <div>Home</div>} />
+			</div>
+		);
+	}
+}
+```
```
matt-gadd: I think after this, you then talk about actually registering the routes, as they now know what an Outlet is.

docs/en/routing/basic-usage.md
agubler: At the moment it's:

1) Route Config
2) Registering the router
3) Using the Outlet widget

Are you suggesting the registration part before the route config? I ordered it this way because the routing config is required to register the router.

docs/en/routing/basic-usage.md
agubler: Sounds good, thanks.

docs/en/routing/basic-usage.md
```diff
@@ -0,0 +1,490 @@
...
+		outlet: 'home',
+		defaultRoute: true,
+		defaultParams: {
+			page: 'about'
+		}
+	}
+];
+```
+
+## Route Configuration
```
matt-gadd: I don't think we should have a separate section for this should we? It'd be better to tell people about the config as we approach each subject.

sbinge: Even simple features can hide a lot of underlying mechanics 😄

I think everything from here could be split out into a separate file. In my view, `basic-usage` is best kept as simple code illustrations of features, with one-liners adding some extra context around the code. Anything more than that - API descriptions, drilling into full-functionality examples, multi-paragraph descriptions, etc - I think should live as additional sections in the reference guide.

docs/en/routing/basic-usage.md
```diff
@@ -0,0 +1,490 @@
...
+
+The routing configuration is a hierarchical structure used to describe the entire Dojo application, associating `outlet` ids to a routing path. The routing path can be nested using children which enables building a routing structure that can accurately reflect the requirements of the application.
+
+The routing configuration API is constructed with the following properties:
+
+* `path: string`: The routing path segment to match in the URL.
+* `outlet: string`: The `outlet` id used to render widgets to the associated routing path.
+* `defaultRoute: boolean`(optional): Marks the outlet as default, the application will redirect to this route automatically if an unknown routes is detected on application load.
```
sbinge: So here's the description of what a default route is - might be worth calling this out & linking to it from the basic usage sample?

docs/en/routing/basic-usage.md
matt-gadd: You could switch it around to be:
* Use an Outlet in a widget
* Create a route
* Register the router

docs/en/routing/basic-usage.md
agubler: Extra routing configuration features are gradually introduced as required for each subject, I still think it's valuable to have a section dedicated to the config.

matt-gadd: What does the extra section give us?

matt-gadd: It's giving me a table of things that given reading the above is absolutely redundant

docs/en/routing/basic-usage.md
```diff
@@ -0,0 +1,490 @@
...
+
+const r = renderer(() => <App />);
+r.mount({ registry });
+```
+
+These history managers work like adapters, meaning that custom history managers can be implemented by fulfilling the history manager interface.
+
+## Error Outlet
+
+A special `outlet` called, `errorOutlet` is registered for that will match when the route doesn't match (`exact` or `partial`) any outlet in the routing configuration. You can use this `outlet` to render a widget to inform the user that the route does not exist.
```
sbinge: How do error outlets and default routes interact?

docs/en/routing/basic-usage.md
matt-gadd: Even re-revisiting the routes file wouldn't be bad:
* Register the router
* Use an Outlet in a widget
* Create a routes file

docs/en/routing/basic-usage.md
agubler: It gives a summary of the routing configuration, I don't think it's redundant to have an overview section.

matt-gadd: okay, good for a separate file as already said then. I read this top down as naively as possible and it seemed redundant

agubler: Cool, I started with everything in the same file - so haven't started to split this out yet. Mostly wanted to get some feed back on the words/examples etc 👍

docs/en/routing/basic-usage.md
matt-gadd: Agreed, this should show an example of using query params like the params above.

docs/en/routing/basic-usage.md
agubler: I'll add some notes, on the default route only applies to the initial application load. If there is a default route it will redirect to it, otherwise when using the application the default route doesn't have any affect.

docs/en/routing/basic-usage.md
```diff
@@ -0,0 +1,490 @@
...
+
+This example would register the following routes and outlets:
+
+| Route             | Outlet          |
+| ----------------- | --------------- |
+| `/home`           | `home`          |
+| `/about`          | `about-overview`|
+| `/about/company`  | `about-company` |
+| `/about/history`  | `about-history` |
+| `/about/{param}`  | `about-services`|
```
matt-gadd: the param bit confuses me here, why is it not `services`?

docs/en/routing/basic-usage.md
```diff
@@ -0,0 +1,490 @@
...
+
+export default class App extends WidgetBase {
+	protected render() {
+		return (
+			<div>
+				<Outlet id="home" renderer={(matchDetails) => {
+					const { params, queryParams, isExact, isError, router } = matchDetails;
+
+					const gotoHome = () => {
+						const link = router.link('home');
```
matt-gadd: what is this example meant to be showing? both `router.link` and `router.setPath` have no explanations or references in this document

docs/en/routing/basic-usage.md
agubler: It was to demonstrate that any other value would match the `about-services` outlet, but I can clean that up and just add a couple of extra paths that both route to `about-services`

docs/en/routing/basic-usage.md
agubler: That's a good point, still need to talk about the router API and then how to use it - this can probably be part of that section.

docs/en/routing/basic-usage.md
matt-gadd: yeah this table doesn't really make sense to me.

docs/en/routing/basic-usage.md
```diff
@@ -0,0 +1,490 @@
...
+				}} />
+			</div>
+		);
+	}
+}
+```
+
+## `Link` and `ActiveLink`
+
+The `Link` component is a wrapper around an `a` DOM element that enables consumers to specify an `outlet` to create a link to. It is also possible to use a static route by setting the `isOutlet` property to `false`.
```
matt-gadd: prefer "anchor tag" over "`a` DOM element"

docs/en/routing/basic-usage.md
agubler: It is supposed to show how a routing paths map to outlets.

docs/en/routing/basic-usage.md
```diff
@@ -0,0 +1,490 @@
...
+import WidgetBase from '@dojo/framework/widget-core/WidgetBase';
+import { tsx } from '@dojo/framework/widget-core/tsx';
+import { Link } from '@dojo/framework/routing/Link';
+
+export default class App extends WidgetBase {
+	protected render() {
+		return (
+			<div>
+				<Link to="home" params={{ foo: 'bar' }}>Link Text</Link>
+				<Link to="#/static-route" isOutlet={false}>Link Text</Link>
```
matt-gadd: should the hash be here? doesn't this make it coupled to the history?

docs/en/routing/basic-usage.md
agubler: lol i wrote anchor tag and changed it, not sure why - noted.

docs/en/routing/basic-usage.md
matt-gadd: yeah like i say, i don't know what the left column is, is it the url? it just says route, but the route had `{services}`. real examples always help (ie the input).

docs/en/routing/basic-usage.md
agubler: You can use `Link` to set a static `href` and not generate the link via the router based on an outlet... So if you did it like this then it would need to have the `#` prefix, however I'm not sure I see the value in supporting this with `Link`.

docs/en/routing/basic-usage.md
matt-gadd: > Marks the outlet as default, the application will redirect to this route automatically if an unknown routes is detected on application load.

Doesn't it also redirect if no route provided?

docs/en/routing/basic-usage.md
sbinge: I think these tables add value as there are multiple ways to approach the guide; linear or ad-hoc. Having a complete reference in one place helps the latter usage.

docs/en/routing/basic-usage.md
agubler: gotcha, I'll make some changes and see if it makes more sense.

docs/en/routing/basic-usage.md
agubler: Correct, I'll add that @matt-gadd specifically - technically no route is an unknown route but it's not obvious.

docs/en/routing/basic-usage.md
matt-gadd: Yeah this functionality doesn't even make sense to me if it doesn't at least decouple the history. This gives me no advantage over an anchor tag

agubler: Aye also my thoughts, I'll remove this from the ref guide.

agubler: As it's confusing if nothing else.

docs/en/routing/basic-usage.md
```diff
@@ -0,0 +1,490 @@
...
+];
+```
+
+>src/App.tsx
+```tsx
+import WidgetBase from '@dojo/framework/widget-core/WidgetBase';
+import { tsx } from '@dojo/framework/widget-core/tsx';
+import Outlet from '@dojo/framework/routing/Outlet';
+
+export default class App extends WidgetBase {
```
matt-gadd: I think it would be better to talk about the different match details functionality separately rather than one large example.

docs/en/routing/basic-usage.md
agubler: I'm not sure about changing the order of these, to me it feels more natural starting with route config, adding routing to the application, then doing something that works with the registered outlet.

matt-gadd: when i read it i'm immediately confused by you saying put an outlet in there with no explanation what an outlet is. if you're going to keep the order, then drop the outlet part of the definition and revisit after you'd explained an outlet?

agubler: I wish it was just `id` in the config

matt-gadd: Nah, that'd be confusing without any qualification in the config because an `id` is presumed unique.

matt-gadd: If you want to keep as is maybe just explicitly say ignoring the outlet for now, we will explore it in a later section or something along those lines.

docs/en/routing/basic-usage.md
agubler: Added a sentence above the describes the behaviour

docs/en/routing/basic-usage.md
agubler: Done

docs/en/routing/basic-usage.md
agubler: Moved to a separate file

agubler: Done

docs/en/routing/basic-usage.md
agubler: Updated

docs/en/routing/basic-usage.md
agubler: I changed the title to say `Url Path` and used multiple examples that both map to the `about-services` outlet. The changes can be seen in the `route-configuration.md` file.

docs/en/routing/outlet-match-details.md
```diff
@@ -0,0 +1,71 @@
...
+import Outlet from '@dojo/framework/routing/Outlet';
+
+export default class App extends WidgetBase {
+	protected render() {
+		return (
+			<div>
+				<Outlet id="home" renderer={(matchDetails) => {
+					const { params, queryParams, isExact, isError, router } = matchDetails;
+
+					const gotoHome = () => {
```
agubler: This comment is still valid: https://github.com/dojo/framework/pull/286#discussion_r263755302

docs/en/routing/basic-usage.md
agubler: Done

docs/en/routing/basic-usage.md
agubler: Removed

docs/en/routing/outlet-match-details.md
```diff
@@ -0,0 +1,71 @@
...
+For every `outlet` that is match on a route change `MatchDetails` are injected into the `Outlet` widget's `renderer` property. The `MatchDetails` object contains specific details for the matched outlet.
+
+* `queryParams: { [index: string]: string }`: The query params from the matched routing.
+* `params: { [index: string]: string }`: The path params from the matched routing.
+* `isExact(): boolean`: A function indicates if the outlet is an exact match for the path.
+* `isError(): boolean`: A function indicates if the outlet is an error match for the path.
+* `type: 'index' | 'partial' | 'error'`: The type of match for the route, either `index`, `partial` or `error`.
+* `router: RouterInterface`: The router instance which can used to create links and initiate route changes.
+
+>src/routes.ts
```
agubler: This comment is still valid: https://github.com/dojo/framework/pull/286#discussion_r263757837

docs/en/routing/history-managers.md
```diff
@@ -0,0 +1,89 @@
...
+
+## StateHistory
+
+`StateHistory` uses the browser's [history API](https://developer.mozilla.org/en-US/docs/Web/API/History), to manage application route changes.
+
+**Note:** The `StateHistory` manager will require server-side machinery to enable an application to support refreshing on a route.
+
+```ts
+import { Router } from '@dojo/framework/routing/Router';
+import { MemoryHistory } from '@dojo/framework/routing/history/StateHistory';
```
xiaohulu: Should be `StateHistory`
```suggestion
import { StateHistory } from '@dojo/framework/routing/history/StateHistory';
```

docs/en/routing/history-managers.md
```diff
@@ -0,0 +1,89 @@
...
+import { Router } from '@dojo/framework/routing/Router';
+
+const router = new Router(config);
+```
+
+## StateHistory
+
+`StateHistory` uses the browser's [history API](https://developer.mozilla.org/en-US/docs/Web/API/History), to manage application route changes.
+
+**Note:** The `StateHistory` manager will require server-side machinery to enable an application to support refreshing on a route.
```
xiaohulu: Should tell what the server-side machinery is. For example:

1. modify the request path to load the static resource （js,css etc）
2. modify the request path to let the dojo routh redirct to home (SPA)

matt-gadd: we can also mention here that dojo build app's --serve mode has this machinery

PR #259, opened 2019-02-15
`ThemeSwitcher`, `LocaleSwitcher` and generic `Provider` widgets

agubler:
**Type:** feature

The following has been addressed in the PR:

* [x] There is a related issue
* [x] All code has been formatted with [`prettier`](https://prettier.io/) as per the [readme code style guidelines](./../#code-style)
* [x] Unit or Functional tests are included in the PR

**Description:**

This pull requests adds a collection of similar widgets that can be used to work with injector values from an applications registry.

* ThemeSwitcher - Injects a function that can be used to change an the theme set in the theme injector
* LocaleSwitcher - Injects a function that can be used to change an the locale data set in the i18n injector
* Provider - A generic provider that enables consumers to work with injectors.

Resolves #258 
Resolves #257 
Resolves #256

PR #300, opened 2019-03-19
Issue #299: Styling & theming reference guide

### LINKED ISSUE #299: Reference guide: styling & theming
As part of our new content initiative, write a reference guide for styling & theming following the established reference guide format.

Existing theming-related content to be consolidated & expanded upon:
* https://github.com/dojo/framework/blob/master/src/widget-core/README.md#styling--theming
* https://github.com/dojo/dojo.io/blob/master/site/source/tutorials/003_creating_widgets/index.md#styling-widgets-with-css-modules
* https://github.com/dojo/dojo.io/blob/master/site/source/tutorials/007_theming/index.md
* https://github.com/dojo/themes/blob/master/README.md
* https://github.com/dojo/cli-create-theme/blob/master/README.md
* https://github.com/dojo/cli-build-theme/blob/master/README.md

sbinge:
**Type:** feature

**Description:** Adds a new styling & theming reference guide, consolidating & expanding upon all current styling and theming related documentation content.

Resolves #299


sbinge:
Note that this includes documentation on the new `ThemeSwitcher` utility widget that will be introduced in https://github.com/dojo/framework/pull/259.

agubler:
In the basic usage section I see that you are referencing the supplemental sections that provide in depth information, but do you think we should at least give a highlight level overview about what the example is demonstrating. If I am in the basic usage, I don't think I'd want to keep jumping around to read in more depth about what the example is showing.

docs/en/styling-and-theming/supplemental.md
```diff
@@ -0,0 +1,642 @@
...
+
+These obfuscated class names are localized to `MyWidget` elements, and are determined by Dojo's CSS modules build process. With this mechanism it is possible for other widgets in the same application to also use the `myWidgetClass` class name with different styling rules, and not encounter any conflicts between each set of styles.
+
+**Warning:** The obfuscated CSS class names should be considered unreliable and may change with a new build of an application, so developers should not explicitly reference them (for examples if attempting to target an element from elsewhere in an application).
+
+## Abstracting and extending stylesheets
+
+### CSS custom properties
+
+Dojo allows use of CSSnext features such as [custom properties and `var()`](http://cssnext.io/features/#custom-properties-var) to help abstract and centralize common styling properties within an application.
```
agubler: We don't use cssnext anymore as it is deprecated, we use https://preset-env.cssdb.org/.

docs/en/styling-and-theming/supplemental.md
```diff
@@ -0,0 +1,642 @@
...
+
+Rather than having to specify the same values for colors or fonts in every widget's CSS module, abstract custom properties can instead be referenced by name, with values then provided in a centralized CSS `:root` pseudo-class. This separation allows for much simpler maintenance of common styling concerns across an entire application.
+
+Note that the `:root` pseudo-class is global within a webpage, but through Dojo's use of CSS modules, `:root` properties could potentially be specified in many locations across an application. However, Dojo does not guarantee the order in which CSS modules are processed, so to ensure consistency of which properties appear in `:root`, it is recommended that applications use a single `:root` definition within a centralized `variables.css` file in the application's codebase. This centralized variables file is a regular CSS file (not a CSS module) and can be `@import`ed as such in any CSS modules that require custom property values.
```
agubler: Do you think we should talk about the fact that by default for evergreen browsers css vars are not resolved at build time, but if building in legacy mode (i.e. need to support IE) that the styles are duplicated, one with the var reference and one with a resolved value determined at build time which IE will fallback to?

docs/en/styling-and-theming/supplemental.md
```diff
@@ -0,0 +1,642 @@
...
+}
+```
+
+> src/themes/myDarkTheme/MyWidget.m.css
+
+```css
+@import '../variables.css';
+
+.root {
+	font-family: var(--font);
```
agubler: I think we had an issue with using a var for `font-family` in IE very recently

docs/en/styling-and-theming/supplemental.md
```diff
@@ -0,0 +1,642 @@
...
+
+[CSS module files](https://github.com/css-modules/css-modules) in Dojo applications can leverage `composes:` functionality to apply sets of styles from one class selector to another. This can be useful when creating a new theme that tweaks an existing one, as well as for general abstraction of common sets of styling properties within a single theme (note that [CSS custom properties](#css-custom-properties) are a more standardized way of abstracting values for individual style properties).
+
+**Warning:** Use of `composes:` can prove brittle, for example when extending a third-party theme that is not under direct control of the current application. Any change made by a third-party could break an application theme that `composes` the underlying theme, and such breakages can be problematic to pin down and resolve.
+
+However, careful use of this feature can be helpful in large applications. For example, centralizing a common set of properties:
+
+> src/themes/common/ButtonArchetype.m.css
```
agubler: Shall we use `base` over `archetype`? I just think it will be more obvious to a wider audience.

docs/en/styling-and-theming/supplemental.md
```diff
@@ -0,0 +1,642 @@
...
+
+Here, the name of the widget is the same as the the name of its CSS module file, but developers should be careful not to mistake the widget's theme key as representing the widget's TypeScript class name.
+
+For a second widget that does not follow CSS module naming conventions, such as `src/widgets/BespokeWidget.ts` that uses a corresponding CSS module such as `src/styles/BespokeStyleSheet.m.css`, its widget theme key would instead be:
+
+    my-app/BespokeStyleSheet
+
+## Writing a theme
+
+Themes are TypeScript modules that export an unnamed default object which maps [widget theme keys](#widget-theme-keys) to typed CSS module imports. CSS modules in a theme are the same as regular modules used directly in widgets. Once a theme is applied in an application, each widget identified via its theme key in the theme's definition object will have its styles overridden with those specified in the CSS module associated with that widget's theme key.
```
agubler: "export an unnamed default object" -> "export a default object"?

docs/en/styling-and-theming/supplemental.md
```diff
@@ -0,0 +1,642 @@
...
+
+There are three requirements for widgets to be considered themeable:
+
+1.  The widget's class should have the `@theme()` decorator applied, passing in the widget's imported CSS module as a decorator argument.
+2.  The widget's parent class should be wrapped with `ThemedMixin`.
+3.  One or more of the widget's styling classes should be wrapped in a call to `this.theme()` when rendering the widget.
+
+By convention, there is a fourth requirement that is useful when developing widgets intended for distribution (this is a convention that widgets in Dojo's widget library follow):
+
+4.  The widget's main styling class should be named `root`. Doing so provides a predictable way to target the main styling class of any third-party themeable widgets when overriding styles in a custom theme.
```
agubler: Do we ever describe that the `root` class should be used on the parent node returned by the widget? Or at least is the general pattern widgets does.

docs/en/styling-and-theming/supplemental.md
```diff
@@ -0,0 +1,642 @@
...
+				<ComboBox classes={myExtraClasses} results={['foo', 'bar']} />
+			</div>
+		);
+	}
+}
+```
+
+Note that it is a widget author's responsibility to explicitly pass the `classes` property to all child widgets that are leveraged, as the property will not be injected nor automatically passed to children by Dojo itself.
+
+## Making themeable applications
```
agubler: Do we think this should be higher up in the order? This feels like it's the first thing that people will want to do.

docs/en/styling-and-theming/basic-usage.md
```diff
@@ -0,0 +1,177 @@
+# Basic Usage
```
sbinge: @agubler referencing [your comment on the main PR](https://github.com/dojo/framework/pull/300#issuecomment-477884342) - do you see some of these specifically as not providing enough detail in their opening sentences? I agree there should be some context around the examples, but I also see the sample code itself as (hopefully) self-descriptive enough to make the usage fully understandable, together with the introductory sentence. The links to the supplemental guide were only really intended for quick navigation to specific areas of interest if a user wants to go into full detail for a feature - not so that they need to read up on something before understanding the sample usage. Happy to add more context to some of these if they're not clear enough as-is - and perhaps we need a better way of linking to supplemental for the above purpose? Maybe a 'See here for more detail' link at the end of the sample code?

docs/en/styling-and-theming/basic-usage.md
```diff
@@ -0,0 +1,177 @@
...
+import * as css from '../styles/MyWidget.m.css';
+
+export default class MyWidget extends WidgetBase {
+	protected render() {
+		return <div classes={[css.root]}>My Widget</div>;
+	}
+}
+```
+
+## Making a widget themeable
```
agubler: @sbinge For example, I'm making the widget themeable which is good. However I have no idea which parts of it are important at first glance - I need to deep dive into the supplemental section....

docs/en/styling-and-theming/supplemental.md
```diff
@@ -0,0 +1,642 @@
...
+
+Dojo widgets function best as simple components that each handle a single responsibility. They should be as encapsulated and modular as possible to promote reusability while avoiding conflicts with other widgets the application may also be using.
+
+Widgets can be styled via regular CSS, but to support encapsulation and reuse goals, each widget should maintain its own independent CSS module that lives parallel to the widget's source code. This allows widgets to be styled independently, without clashing on similar class names used elsewhere in an application.
+
+Dojo differentiates between two types of styling, each representing a different granularity of styling concerns within an enterprise web application:
+
+-   [Structural styles](#structural-widget-styling) (_granularity:_ per-widget)
+    -   The minimum styles necessary for a widget to function. These are localized to the particular widget they are needed for.
+-   [Presentational styles](#theming-a-dojo-application) (_granularity:_ application-wide)
```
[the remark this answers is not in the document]
sbinge: Is there actually a third category in that case? 'Themeable styles'?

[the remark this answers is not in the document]
agubler: Themeable styles are still per widget and not generally shared across different widget

docs/en/styling-and-theming/supplemental.md
```diff
@@ -0,0 +1,642 @@
...
+    -   Styles - usually visual - that may be changed without impacting widget functionality. These can cut across several widgets and usually provide a consistent theme for all widgets used within an application.
+
+## Structural widget styling
+
+Dojo leverages [CSS Modules](https://github.com/css-modules/css-modules) to provide all of the flexibility of CSS, but with the additional benefit of localized classes to help prevent inadvertent styling collisions across a large application. Dojo also makes use of [typed CSS modules](https://github.com/Quramy/typed-css-modules), allowing widgets to import their CSS modules similar to any other TypeScript module, and refer to specific CSS class names at design-time via IDE autocompletion.
```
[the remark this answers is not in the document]
sbinge: 👍 this was a holdover from the original README.

docs/en/styling-and-theming/supplemental.md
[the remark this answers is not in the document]
agubler: @tomdye Can probably clarify this better the me (or tell me I'm wrong of course)

docs/en/styling-and-theming/supplemental.md
agubler: @rorticus @tomdye do you remember this? An issue I think was fixed in themes for this?

docs/en/styling-and-theming/supplemental.md
sbinge: I originally had it near the top, but decided to be consistent with the ordering we established for i18n, i.e. first set up all the pieces, then plug them together.

That said, theming may be different in that while there are no 'default' message bundles, apps could use an out-the-box theme without needing to author anything themselves.

Would expected usage here override a desire for consistency across documentation?

docs/en/styling-and-theming/supplemental.md
sbinge: Is it worth just removing `font-family` from the examples?

docs/en/styling-and-theming/basic-usage.md
sbinge: Actually, within `basic-usage` I've been assuming a linear reading order, where each example builds upon the previous. Users should hopefully notice the difference between the original `MyWidget` above and the modifications done to it for this example to make it themeable.

But, that intention is not obvious, and breaks down for ad-hoc reading. Unless we have a better way of presenting diffs between examples, it may be worth making each example more standalone as you describe?

docs/en/styling-and-theming/introduction.md
```diff
@@ -0,0 +1,13 @@
...
+| Per-widget stylesheets                | [CSS Modules](https://github.com/css-modules/css-modules) can be used to define stylesheets that are scoped to individual widgets, avoiding potential cross-contamination and style name clashes. A widget can refer to its exact CSS class names via [typed CSS module imports](https://github.com/Quramy/typed-css-modules) and IDE autocompletion.                                                  |
+| Abstracted CSS properties             | Individual CSS modules can refer to a set of centrally-defined `:root` style variables through [CSS custom properties and `var()`](http://cssnext.io/features/#custom-properties-var).                                                                                                                                                                                                                 |
```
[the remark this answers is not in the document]
tomdye: agreed re css-next, it's been deprecated and we no longer use it.

docs/en/styling-and-theming/basic-usage.md
```diff
@@ -0,0 +1,177 @@
...
+import myTheme from '../themes/MyTheme/theme';
+import alternativeTheme from '../themes/MyAlternativeTheme/theme';
+
+import * as css from '../styles/App.m.css';
+
+@theme(css)
+class App extends ThemedMixin(WidgetBase) {
+	protected render() {
+		return (
+			<div classes={[this.theme(css.container)]}>
```
tomdye: perhaps drop the wrapping divs to make this a bit smaller, the formatting is a little off as well. could just be GHs big spaces though

docs/en/styling-and-theming/introduction.md
```diff
@@ -0,0 +1,13 @@
+# Introduction
+
+As any HTML-based technology, Dojo makes use of CSS for styling elements across the framework and for applications developed with it.
```
tomdye: I'd remove the first part, as others dont use css

docs/en/styling-and-theming/basic-usage.md
sbinge: This is all prettier auto-formatted at the moment!

docs/en/styling-and-theming/supplemental.md
```diff
@@ -0,0 +1,642 @@
+# Styling & Theming in Dojo
+
+Dojo widgets function best as simple components that each handle a single responsibility. They should be as encapsulated and modular as possible to promote reusability while avoiding conflicts with other widgets the application may also be using.
+
+Widgets can be styled via regular CSS, but to support encapsulation and reuse goals, each widget should maintain its own independent CSS module that lives parallel to the widget's source code. This allows widgets to be styled independently, without clashing on similar class names used elsewhere in an application.
```
tomdye: it provides obfuscation and encapsulation as well. Stops styles from leaking.

docs/en/styling-and-theming/supplemental.md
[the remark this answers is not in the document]
tomdye: it's upto the widget dev which css they pass via `this.theme` and which they do not. The classes not passed to `this.theme` would be things like laying out children etc. 90% of classes should be passed via `this.theme`. The third category might be shared styles that get imported to multiple widgets but this is more of an APP concern than a widget library concern. An example of this could be a typography.m.css file that provides `bold` / `header` classes to be re-used across an app.

docs/en/styling-and-theming/supplemental.md
```diff
@@ -0,0 +1,642 @@
...
+-   [Structural styles](#structural-widget-styling) (_granularity:_ per-widget)
+    -   The minimum styles necessary for a widget to function. These are localized to the particular widget they are needed for.
+-   [Presentational styles](#theming-a-dojo-application) (_granularity:_ application-wide)
+    -   Styles - usually visual - that may be changed without impacting widget functionality. These can cut across several widgets and usually provide a consistent theme for all widgets used within an application.
+
+## Structural widget styling
+
+Dojo leverages [CSS Modules](https://github.com/css-modules/css-modules) to provide all of the flexibility of CSS, but with the additional benefit of localized classes to help prevent inadvertent styling collisions across a large application. Dojo also makes use of [typed CSS modules](https://github.com/Quramy/typed-css-modules), allowing widgets to import their CSS modules similar to any other TypeScript module, and refer to specific CSS class names at design-time via IDE autocompletion.
+
+The CSS module file for a widget should have a `.m.css` extension, and by convention is usually named the same as the widget it is associated with.
```
tomdye: is it worth noting that `.m.css` files get processed differently?

docs/en/styling-and-theming/supplemental.md
```diff
@@ -0,0 +1,642 @@
...
+import * as css from '../styles/MyWidget.m.css';
+
+export default class MyTsxWidget extends WidgetBase {
+	protected render() {
+		return <div classes={[css.myWidgetClass, css.myWidgetExtraClass]}>Hello from a Dojo TSX widget!</div>;
+	}
+}
+```
+
+When inspecting the CSS classes of these sample widgets in a built application, they will not contain `myWidgetClass` and `myWidgetExtraClass`, but rather obfuscated CSS class names similar to `MyWidget-m__myWidgetClass__33zN8` and `MyWidget-m__myWidgetExtraClass___g3St`.
```
tomdye: I believe we might just use an 8 char hash now?

docs/en/styling-and-theming/supplemental.md
tomdye: yes

docs/en/styling-and-theming/supplemental.md
agubler: Nope we don't, we use a more descriptive hash for both dev and dist

docs/en/styling-and-theming/introduction.md
```diff
@@ -0,0 +1,13 @@
+# Introduction
+
+As a HTML-based technology, Dojo makes use of CSS for styling elements across the framework and for applications developed with it.
```
msssk: "an" before "HTML" (ref: [APA style](https://blog.apastyle.org/apastyle/2012/04/using-a-or-an-with-acronyms-and-abbreviations.html))

msssk: Ohhhh... I just remembered which side of the pond you're on... up to you I guess!

docs/en/styling-and-theming/introduction.md
```diff
@@ -0,0 +1,13 @@
...
+
+Dojo promotes encapsulated structural styling of individual widgets for maximum reuse, as well as simplified presentational theming across all widgets within an application. This pattern gives users a predictable way to style and theme their applications, even when using a mixture of widgets from [Dojo's `@dojo/widgets` library](https://github.com/dojo/widgets), a third-party provider, or any that may be developed in-house for a particular application.
+
+| Feature                               | Description                                                                                                                                                                                                                                                                                                                                                                                            |
```
msssk: The content in this table seems more like a series of concepts than tabular data, and might be better presented as a series of headings with paragraphs.

docs/en/styling-and-theming/basic-usage.md
```diff
@@ -0,0 +1,177 @@
...
+export default class MyWidget extends WidgetBase {
+	protected render() {
+		return <div classes={[css.root]}>My Widget</div>;
+	}
+}
+```
+
+## Making a widget themeable
+
+Extending `ThemedMixin`, applying the `@theme()` decorator to a widget's imported CSS modules, and wrapping CSS class names in calls to `this.theme()` when rendering to allow a [widget's default styles to be overridden by a theme](./supplemental.md#making-themeable-widgets):
```
msssk: This sentence is pretty dense and can be difficult to follow. Might be better as a list?
* extend `ThemedMixin`
* apply `@theme()` decorator
* use `this.theme()`

docs/en/styling-and-theming/basic-usage.md
msssk: I think stating up-front that the examples build on one another would help clarify this.

docs/en/styling-and-theming/supplemental.md
```diff
@@ -0,0 +1,763 @@
...
+
+If using an [independently-distributed theme](#distributing-themes) in its entirety, applications will also need to integrate the theme's overarching `index.css` file into their own styling. This can be done via an import in the project's `main.css` file:
+
+> src/main.css
+
+```css
+@import '@{myThemePackageName}/{myThemeName}/index.css';
+```
+
+By contrast, another way of using only portions of an eternally-built theme within a newly-created application theme is [via theme composition](#composing-off-dojo-themes).
```
msssk: "externally-built"?

docs/en/styling-and-theming/supplemental.md
```diff
@@ -0,0 +1,763 @@
...
+}
+```
+
+## Making themeable widgets
+
+There are three requirements for widgets to be considered themeable:
+
+1.  The widget's class should have the `@theme()` decorator applied, passing in the widget's imported CSS module as a decorator argument.
+    -   Note that widgets can import and reference multiple CSS modules, in which case additional `@theme()` decorators can be stacked, one for each corresponding CSS import intended to be themeable.
+2.  The widget's parent class should be wrapped with `ThemedMixin`.
```
msssk: Is this accurate? Is it not simply that the widget's class should extend `ThemedMixin`?

docs/en/styling-and-theming/supplemental.md
```diff
@@ -0,0 +1,763 @@
...
+    -   described in the [Passing extra classes to widgets](#passing-extra-classes-to-widgets) section.
+
+### Themeable Widget Example
+
+Given the following CSS module file for a themeable widget:
+
+> src/styles/MyThemeableWidget.m.css
+
+```css
+/* requirement 4, i.e. this widget is intended for wider distribution and therefore the main presentational styles for
```
msssk: Manually wrapping this line at (or before?) "presentational" would reduce the likelihood of causing horizontal scrolling.

docs/en/styling-and-theming/supplemental.md
```diff
@@ -0,0 +1,763 @@
...
+				Hello from a themed Dojo TSX widget!
+			</div>
+		);
+	}
+}
+```
+
+#### Using several CSS modules
+
+Widgets can also import and reference multiple CSS modules - this provides another way to abstract and reuse common styling properties through typescript code, in addition to the CSS-based methods described elsewhere in this guide ([CSS custom properties](#css-custom-properties) and [CSS module composition](#css-module-composition)).
```
msssk: typescript -> TypeScript

docs/en/styling-and-theming/supplemental.md
```diff
@@ -0,0 +1,763 @@
...
+				Hello from a themed Dojo TSX widget!
+			</div>
+		);
+	}
+}
+```
+
+### Overriding the theme of specific widget instances
+
+Users of a widget can override the theme of a specific instance by passing in a [valid theme](#writing-a-theme) to the instance's [`theme` widget property](#themedmixin-widget-properties). This is useful when needing to display a given widget in multiple ways across several occurrences within an application.
```
msssk: "`theme` widget property" might be clearer as just "`theme` property"

docs/en/styling-and-theming/supplemental.md
```diff
@@ -0,0 +1,763 @@
...
+            │   {widget}.m.css
+            │   {widget}.m.css.js
+
+## Distributing themes
+
+Dojo's [`cli-build-theme`](https://github.com/dojo/cli-build-theme) package provides a CLI command to help build themes that are intended for distribution across multiple applications. It will create all files necessary to [use the theme in a variety of different ways](#theming-a-dojo-application).
+
+Note that when using [`dojo create theme`](#scaffolding-themes-for-third-party-widgets) to scaffold a new theme, there is no need to use `dojo build theme`, as all relevant files will already be in place. This applies to themes in projects that are built either via [`@dojo/cli-build-app`](https://github.com/dojo/cli-build-app) or [`@dojo/cli-build-widget`](https://github.com/dojo/cli-build-widget).
+
+To use the tooling, install `@dojo/cli-build-theme` locally in a themes project:
```
msssk: Is this correct, or should it be themed? theme?

> a themed project

or 

> a theme project

docs/en/styling-and-theming/supplemental.md
```diff
@@ -0,0 +1,763 @@
...
+
+1.  To use the themes, install `@dojo/themes` into your project, for example through `npm i @dojo/themes`. Then, for regular Dojo applications:
+
+2.  Import the theme CSS into your project's `main.css`:
+
+    ```css
+    @import '~@dojo/themes/dojo/index.css';
+    ```
+
+3.  Import the theme typescript module and use it [as per any other theme](#making-themeable-applications):
```
msssk: TypeScript

docs/en/styling-and-theming/basic-usage.md
sbinge: Added a note to clarify the ordering

docs/en/styling-and-theming/basic-usage.md
sbinge: Added a note at the top of the file to clarify this.

docs/en/styling-and-theming/introduction.md
sbinge: This is just the introduction section though, and follows the convention established in the `i18n` guide. It's only really intended as an elevator pitch, highlighting the main features/selling points of the capability - ideally as one-liner descriptions.

I think the ideal presentation for these is in a description list, but markdown doesn't easily support those without explicit HTML tags unfortunately! Tables felt like the best compromise given all the above.

sbinge:
> A few comments/questions, plus a couple more observations:
> 
> 1. There is no example of passing through an explicit theme to a widget, to override the application theme.
> 2. We don't mention the ability to "stack" the `@theme` uses, for example:
> 
> ```ts
> @theme(myCommonCss)
> @theme(myCss)
> class MyWidget extends WidgetBase {
>     // stuff
> }
> ```

Examples for both the above have been added.

docs/en/styling-and-theming/supplemental.md
```diff
@@ -141,229 +151,148 @@ However, careful use of this feature can be helpful in large applications. For e
...
-
-Here, the name of the widget is the same as the the name of its CSS module file, but developers should be careful not to mistake the widget's theme key as representing the widget's TypeScript class name.
+## Dojo styling best-practices
 
-For a second widget that does not follow CSS module naming conventions, such as `src/widgets/BespokeWidget.ts` that uses a corresponding CSS module such as `src/styles/BespokeStyleSheet.m.css`, its widget theme key would instead be:
+As styles in a Dojo application are mostly scoped to individual widgets, there is little need for complex selector targeting. Style application in Dojo should be as simple as possible - developers can achieve this by following a few simple recommendations:
 
-    my-app/BespokeStyleSheet
+-   Maintain encapsulated widget styling
+    -   Only refer to classes relevant for the corresponding widget within a given CSS module
```
tomdye: not sure this is entirely true, something along the lines of single css module per concern with the exception of intended shared modules, ie an app wide typography module

PR #301, opened 2019-03-19
Correct links to the i18n guide now there's no longer an index.md

sbinge:
**Type:** bug

**Description:** With the updated reference guide format, sans `index.md` file, links from elsewhere in the framework documentation should route to the general reference guide folder for the moment.

**Note** that there are also other minor formatting changes included in this PR as a result of prettier re-writing affected markdown files pre-commit.


agubler:
Ah the pattern for the markdown formatting doesn't catch the root readme... Will fix that.

PR #306, opened 2019-03-22
Merges widget-core into core.

### LINKED ISSUE #280: Moving widget-core to core
**Enhancement**

When we started working on `widget-core`, we wanted to call it `core` but unfortunately `core` already existed as a lot of modules transferred over from the initial `dojo` to `dojo2` transition. Given since then we have effectively removed all those legacy things and we have a total 4 utility modules left in `core`, it's difficult to see why this would exist anymore on it's own. In which case we could move those 4 utility modules over to `widget-core`, and rename `widget-core` to `core`

agubler:
**Type:** feature

**Description:**

Merges all the `widget-core` modules into `core`.

Resolves #280


agubler:
Will redo from the latest master

PR #314, opened 2019-04-08
Assertion Template Enhancements

matt-gadd:
**Type:** feature

**Description:**
TODO: Fill this out explaining enhancements.

API's changed:
```
replace -> replaceChildren
```
(this https://github.com/dojo/framework/pull/247#discussion_r256571684 came back to haunt me)

API's added:
```
replace
remove
setProperties
getProperties
```

New Assertion Template primitive:
```
Ignore
```

### DIFF
--- a/src/testing/assertionTemplate.ts
+++ b/src/testing/assertionTemplate.ts
@@ -154,13 +154,13 @@ export function assertionTemplate(renderFunc: () => DNode | DNode[]) {
 		const node = guard(findOne(render, selector));
 		return node.children || [];
 	};
-	assertionTemplateResult.replace = (selector: string, node: DNode) => {
+	assertionTemplateResult.replace = (selector: string, newNode: DNode) => {
 		return assertionTemplate(() => {
 			const render = renderFunc();
 			const node = guard(findOne(render, selector));
 			const parent = (node as any).parent;
 			const children = [...parent.children];
-			children.splice(children.indexOf(node), 1, node);
+			children.splice(children.indexOf(node), 1, newNode);
 			parent.children = children;
 			return render;
 		});
--- a/src/testing/assertionTemplate.ts
+++ b/src/testing/assertionTemplate.ts
@@ -160,8 +160,8 @@ export function assertionTemplate(renderFunc: () => DNode | DNode[]) {
 			const node = guard(findOne(render, selector));
 			const parent = (node as any).parent;
 			const children = [...parent.children];
-			children.splice(children.indexOf(node), 1);
-			parent.children = [node, ...children];
+			children.splice(children.indexOf(node), 1, node);
+			parent.children = children;
 			return render;
 		});
 	};
--- a/src/testing/assertionTemplate.ts
+++ b/src/testing/assertionTemplate.ts
@@ -20,6 +20,7 @@ export interface AssertionTemplateResult {
 	getProperty(selector: string, property: string): any;
 	getProperties(selector: string): any;
 	replace(selector: string, node: DNode): AssertionTemplateResult;
+	remove(selector: string): AssertionTemplateResult;
 }
 
 const findOne = (nodes: DNode | DNode[], selector: string): DNode | undefined => {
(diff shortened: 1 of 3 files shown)


aciccarello:
Don't forget to update https://github.com/dojo/framework/blob/master/src/testing/README.md

src/testing/assertionTemplate.ts
```diff
@@ -127,11 +144,38 @@ export function assertionTemplate(renderFunc: () => DNode | DNode[]) {
...
 		return node.children || [];
 	};
+	assertionTemplateResult.replace = (selector: string, node: DNode) => {
+		return assertionTemplate(() => {
+			const render = renderFunc();
+			const node = guard(findOne(render, selector));
+			const parent = (node as any).parent;
+			const children = [...parent.children];
+			children.splice(children.indexOf(node), 1);
+			parent.children = [node, ...children];
```
aciccarello: Wouldn't this always put the replacement node at the beginning of the children? You should be able to pass the replacement with `.splice()`.

src/testing/support/assertRender.ts
```diff
@@ -98,9 +99,67 @@ function formatNode(node: WNode | VNode, tabs: any): string {
...
+	}
+	return [actualDecoratedNodes, expectedDecoratedNodes];
+}
+
 export function assertRender(actual: DNode | DNode[], expected: DNode | DNode[], message?: string): void {
-	const parsedActual = formatDNodes(actual);
-	const parsedExpected = formatDNodes(expected);
+	const [decoratedActual, decoratedExpected] = decorate(actual, expected);
+	const parsedActual = formatDNodes(Array.isArray(actual) ? decoratedActual : decoratedActual[0]);
+	const parsedExpected = formatDNodes(Array.isArray(expected) ? decoratedExpected : decoratedExpected[0]);
```
aciccarello: Can you explain what is going on with the `isArray` check? Shouldn't this be changed to the following?
```suggestion
	const parsedExpected = formatDNodes(Array.isArray(decoratedExpected) ? decoratedExpected[0] : decoratedExpected);
```

src/testing/support/assertRender.ts
```diff
@@ -98,9 +99,67 @@ function formatNode(node: WNode | VNode, tabs: any): string {
...
+	expected = Array.isArray(expected) ? expected : [expected];
+	let actualDecoratedNodes = [];
+	let expectedDecoratedNodes = [];
+	const length = actual.length > expected.length ? actual.length : expected.length;
+	for (let i = 0; i < length; i++) {
+		let actualNode = actual[i];
+		let expectedNode = expected[i];
+
+		if (isWNode(expectedNode)) {
+			[actualNode, expectedNode] = assertionWidgets.reduce(
```
aciccarello: this part is somewhat confusing to me. I can follow what it is doing, but it isn't clear to me why it i doing what it is doing. If you are changing the `actualNode` and `expectedNode` references, does it really make sense to use `reduce` here? I think a `forEach` might end up being simpler.

src/testing/support/assertRender.ts
matt-gadd: @aciccarello the `decorator` function always turns nodes into an array (if it isn't already) and returns an array. the `assertRender` function and `formatDNodes` take an individual node or an array. so all this does is use the first node if the original nodes (expected and actual) were a singular node. So no the suggested change is incorrect. I might change the decorate function to align with `formatDNodes` to avoid this slightly confusing conversion though.

src/testing/assertionTemplate.ts
matt-gadd: thanks was being silly here from the copy pasta from remove

src/testing/support/assertRender.ts
matt-gadd: I agree, i just removed the loop completely in the end as the indirection wasn't worth it.

aciccarello:
@matt-gadd is there anything I can do to help with this?

agubler:
superseded by #412

PR #317, opened 2019-04-10
We should skip the same URL

xiaohulu:
**Type:** bug

**Description:**

I use `Link` widget to navigate, but If I click the same `Link` n times, I must click Back button n times to back to previous route. We should skip the same url.

### DIFF
--- a/src/routing/history/StateHistory.ts
+++ b/src/routing/history/StateHistory.ts
@@ -53,10 +53,12 @@ export class StateHistory implements HistoryInterface {
 	}
 
 	public set(path: string) {
-		if (path === this._window.location.pathname) {
+		const value = stripBase(this._base, path);
+		if (this._current === value) {
 			return;
 		}
-		this._window.history.pushState({}, '', this.prefix(stripBase(this._base, path)));
+
+		this._window.history.pushState({}, '', this.prefix(value));
 		this._onChange();
 	}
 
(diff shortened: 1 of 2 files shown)


agubler:
@xiaohulu Thank you, do you mind adding a unit test for the change?

xiaohulu:
I want to add a unit test, but I do not know how to test it and I can not run `npm run test:local` sucess 🙁 
image

xiaohulu:
I do not know how to run `StateHistory` unit test.  😢 

I write the unit test code here, I do not know if it can pass.

```ts
	it('update path, skip the same path', () => {
		const history = new StateHistory({ onChange, window: sandbox.contentWindow! });
		history.set('foo');
		history.set('bar');
		history.set('bar');
		sandbox.contentWindow!.history.back();
		assert.equal(history.current, '/foo');
		assert.equal(sandbox.contentWindow!.location.pathname, '/foo');
	});
```

agubler:
@xiaohulu Hmmm that's interesting, I seem to be able to run the tests (on mac, node: v8.9.1 and npm: 6.4.1).

Have you tried, removing and reinstalling node_modules (and removing the package-lock too)?

src/routing/history/StateHistory.ts
```diff
@@ -53,6 +53,9 @@ export class StateHistory implements HistoryInterface {
 	}
 
 	public set(path: string) {
+		if (path === this._window.location.pathname) {
```
agubler: @xiaohulu Do you think we should move the logic from `_onChange` that does a check against the current and previous value to `set` function? 

```ts
const pathName = this._window.location.pathname.replace(/\/$/, '');
const value = stripBase(this._base, pathName + this._window.location.search);
if (this._current === value) {
	return;
}
```

xiaohulu:
@agubler What is the `base` in the framework, and what is its purpose, and why `this._current` is assign to `stripBase` result

agubler:
@xiaohulu Thanks!

PR #320, opened 2019-04-22
Throwing an error on missing selector in test harness trigger

### LINKED ISSUE #318: Test harness should throw on missing selectors
**Enhancement** 
The testing harness should throw when the selector isn't found. Th.is would help debug failing tests

Package Version: 

**Code**
```typescript
h.trigger('@nonExistantSelector', 'whatToTrigger');
```

**Expected behavior:**
The trigger call will throw an error indicating the selector could not be found.

**Actual behavior:**
The trigger passes silently.

rorticus:
**Type:** bug / feature

**Description:**

Instead of silently failing, throwing an error when a `h.trigger` encounters a selector that doesn't return any nodes.

Resolves #318

### DIFF
--- a/src/testing/assertionTemplate.ts
+++ b/src/testing/assertionTemplate.ts
@@ -16,7 +16,7 @@ export interface AssertionTemplateResult {
 	getProperty(selector: string, property: string): any;
 }
 
-const findOne = (nodes: DNode | DNode[], selector: string): NodeWithProperties => {
+const findOne = (nodes: DNode | DNode[], selector: string): DNode | undefined => {
 	let finalSelector = selector;
 	if (selector.indexOf('~') === 0) {
 		finalSelector = `[\\~key='${selector.substr(1)}']`;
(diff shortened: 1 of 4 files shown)


src/testing/harness.ts
```diff
@@ -133,16 +133,19 @@ export function harness(
...
-					triggerFunction = (firstItem.properties as any)[functionSelector];
-				} else {
-					triggerFunction = functionSelector(firstItem);
-				}
-				if (triggerFunction) {
-					return triggerFunction.apply(widget, args);
-				}
+
+			if (!firstItem) {
+				throw new Error(`Cannot find node with selector ${selector}`);
```
aciccarello: I like this error message. It might be good to make this and the assertion template errors consistent.

https://github.com/dojo/framework/blob/4526fecf0feb91df26ad8f3a8f030f13822dff71/src/testing/assertionTemplate.ts#L36

rorticus: I don't disagree, but updating the message there is a bit more complicated. I've made the change though, we'll see how it goes 😄

src/testing/assertionTemplate.ts
```diff
@@ -26,21 +26,19 @@ const findOne = (nodes: DNode | DNode[], selector: string): DNode | undefined =>
...
-
-type NodeWithProperties = (VNode | WNode) & { properties: { [index: string]: any } };
 
-const guard = (node: DNode): NodeWithProperties => {
 	if (!node) {
-		throw Error('Node not found');
+		throw Error(`Cannot find node with selector ${selector}`);
 	}
 	if (!isWNode(node) && !isVNode(node)) {
 		throw Error('Cannot set or get on unknown node');
```
matt-gadd: Hmm I don’t think we should change this. The findOne is meant to be equivalent to querySelector and not throw. It also doesn’t make sense to throw a setting unknown node here given the function name?

rorticus: Hehe, yeah, I wasn't sure about these changes. The other option is i can pass the selector in `guard` so it can use it in the error message, but thats going to end up looking silly and be a bit of a pain every time you call it.  Maybe i should just remove that last commit?

matt-gadd: Ha yeah I agree, maybe for now we just stick with the trigger change and we can raise a separate issue for making error messages more consistent (it might be that we use the same function in the harness and assertion template)

PR #337, opened 2019-05-11
Build reference documentation

### LINKED ISSUE #322: Reference guide: Building Dojo Applications
Covering all aspects of how to build Dojo applications. Merging, updating & streamlining content from:
* Quickly scaffolding new applications: https://github.com/dojo/cli-create-app/blob/master/README.md
* Building applications + difference between dev and production builds: https://github.com/dojo/cli-build-app/blob/master/README.md
* Using bundles - benefits around resource reduction + automatic deferred loading; how to define them; how cross-bundle duplication may occur for shared resources across multiple bundles
* Creating PWAs
* Using BTR within an application
* Using has() for build time elision of ununsed features - https://github.com/dojo/framework/blob/master/src/has/README.md
* Other relevant aspects (dojorc not covered in any of the above; ejection) from https://github.com/dojo/dojo.io/blob/master/site/source/tutorials/006_deploying_to_production/index.md

Info about "locales" and "cldrPaths" can just link to the i18n guide.

devpaul:
Creates a reference guide for building Dojo applications. I tried to strike a balance between providing too much information and just enough to get things going. Basic usage was kept as simple as possible while the supplemental goes deeper. I still held back from providing every single detail even in the supplemental section and tried to focus on what would provide value to someone looking to use a particular feature.

There are a few TODOs in here that couldn't be addressed at the time of writing or are looking for feedback.

This information was pulled together from a ton of different sources. Most of it was tested and validated against a new app. I expect there may be some things I left out that should be added back in. Feedback welcome!

Resolves #322


docs/en/building/basic-usage.md
```diff
@@ -0,0 +1,118 @@
+# Basic Usage
+
+## Creating a Dojo application
```
agubler: Do you think we need to talk about creating an application in the build reference guide?

docs/en/building/basic-usage.md
```diff
@@ -0,0 +1,118 @@
...
+
+```bash
+npm run dev
+```
+
+The application will also have source maps available. This allows the debugger to map the built JavaScript code back to the original TypeScript code that lives in the `src/` directory.
+
+For more information on how to configure the build options see the [@dojo/cli-build-app](https://github.com/dojo/cli-build-app/blob/master/README.md#configuration) README.
+
+## Testing
```
agubler: I don't think we need to talk about testing in detail for the build reference guide over explaining how to build the `unit` and `functional` bundles (with a link out to the testing reference guide, when it's done!)

docs/en/building/supplemental.md
```diff
@@ -0,0 +1,541 @@
...
+
+## Bundling considerations
+
+<!-- TODO I am not confident in what I am saying here. Under what conditions will duplication of common/shared resources occur? How do we avoid this? Can we define a bundle for common widgets? Should I talk about the [bundle analyzer](https://github.com/dojo/cli-build-app/blob/master/README.md#bundle-analyzer) -->
+
+Under the covers Dojo uses Webpack with a number of custom plugins to provide intelligent application bundling. However, sometimes decisions made by the build tool or manually defined in `.dojorc` can create duplication of common resources shared by multiple bundles. Some of this is unavoidable. A good general rule of thumb for avoid duplication is to try to ensure that common code is at the outermost edges of your dependency tree. In other words, minimize dependencies as much as possible among shared code. If a significant amount of code may be shared among bundles (e.g. common widgets) consider bundling these assets together.
```
agubler: We shouldn't need to mention `webpack` at all, it's an implementation detail 😄

docs/en/building/supplemental.md
```diff
@@ -0,0 +1,541 @@
...
+if (true) {
+	console.log('Starting in production');
+} else {
+	console.log('Starting in dev mode');
+}
+
+export const mode = true ? 'dist' : 'dev';
+```
+
+Webpack's dead branch removal would then remove the unreachable code.
```
agubler: This seems to be too low level to explain in a reference guide and it's not actually webpack that does this anyway.

docs/en/building/supplemental.md
```diff
@@ -0,0 +1,541 @@
...
+import 'b'; // removed
+'!has("foo")';
+import 'c';
+'has("baz")';
+import 'd';
+```
+
+In the above example, imports `a` and `b` would be removed because their values evaluate to `true`. Import `c` will remain because it evaluates to `false`. Import `d` will remain because the `baz` feature has not been statically defined.
+
+## Known features
```
agubler: I don't think this is correct anymore, we simply have `modern` - however I am not sure this is something that a user should really need to know.

docs/en/building/supplemental.md
```diff
@@ -0,0 +1,541 @@
...
+| `debug`             | Provides a way to create a code path for code that is only usable when debugging or providing enhanced diagnostics that are not desired in a _production_ build. Defaults to `true` but should be configured statically as `false` in production builds. |
+| `host-browser`      | Determines if the current environment contains a `window` and `document` object in the global context, therefore it is generally safe to assume the code is running in a browser environment.                                                            |
+| `host-node`         | Attempts to detect if the environment appears to be a node environment.                                                                                                                                                                                  |
+| `build-time-render` | Statically defined by the build-time rendering system during build-time rendering.                                                                                                                                                                       |
+
+# Build-time Rendering
```
agubler: This is a flagship feature along with the PWA support, we should move this to the top.

docs/en/building/supplemental.md
```diff
@@ -0,0 +1,541 @@
...
+	protected render() {
+		const message = this.meta(Block).run(readFile)('../content/hello-dojo-blocks.txt');
+		return v('div', [message]);
+	}
+}
+```
+
+This widget runs the `read-file.block.ts` module at build time, loading the file path passed which gets used as the children in the widget.
+
+## Snapshots
```
agubler: This section does not feel necessary... the result of build time rendering should be explain in the main section?

docs/en/building/supplemental.md
```diff
@@ -0,0 +1,541 @@
...
+> Uglify output
+
+```js
+console.log('Starting in production');
+export const mode = 'dist';
+```
+
+Any features which are not statically asserted, are not re-written. This allows the code to determine at run-time if the feature is present.
+
+## Elided Imports
```
agubler: This is a very advanced use case and I would be tempted to not include it in the user facing documentation.

docs/en/building/basic-usage.md
```diff
@@ -0,0 +1,118 @@
...
+## Building
+
+There are several types of build targets or `mode`s that Dojo's CLI tool supports. All of these modes can be seen in the various `package.json` scripts generated by `dojo create app`.
+
+Run the following to create an optimized build ready for production.
+
+> Command line
+
+```bash
+npm run build
```
agubler: Can we use the "real" dojo commands through out, because unless users start with the cli template the changes are they wont have the same npm scripts.

docs/en/building/basic-usage.md
```diff
@@ -0,0 +1,118 @@
...
+
+> Command line
+
+```bash
+npm run dev
+```
+
+The application will also have source maps available. This allows the debugger to map the built JavaScript code back to the original TypeScript code that lives in the `src/` directory.
+
+For more information on how to configure the build options see the [@dojo/cli-build-app](https://github.com/dojo/cli-build-app/blob/master/README.md#configuration) README.
```
agubler: We shouldn't be pointing users back to the README, all the information needed should be included in the reference guides.

docs/en/building/supplemental.md
```diff
@@ -0,0 +1,541 @@
...
+		}
+	}
+}
+```
+
+This configuration describes two routes. A `home` route and a more complex `comments` route. The `comments` route is a more complex route with parameter data. A `match` is used to make sure that the build-time HTML created for this route is applied to any route that matches the regular expression.
+
+BTR generates a screenshot for each of the paths rendered during the build in the output/info/screenshots directory of your project.
+
+### Hash history
```
agubler: This header seems confusing to me.

docs/en/building/supplemental.md
```diff
@@ -0,0 +1,541 @@
...
+| `build-time-render` | Statically defined by the build-time rendering system during build-time rendering.                                                                                                                                                                       |
+
+# Build-time Rendering
+
+BTR renders a route to HTML during the build process and in-lines critical CSS and assets needed to display the initial view. This allows Dojo to pre-render the initial HTML used by a route and inject it directly into the page immediately. Skipping the initial render by creating it during the build process gives us many of the same benefits of server side rendering (SSR) such as performance gains and search engine optimization without the complexities of SSR.
+
+## Using BTR
+
+First make sure `index.html` includes a DOM node with an `id` attribute. This node will be used by Dojo's virtual DOM to compare and render the application's HTML. BTR requires this setup so it can render the HTML generated at build time. This creates a very fast and responsive initial rendering of the route. Once the route has had a change to load and render the virtual DOM is able to diff against the BTR.
```
agubler: "Once the route has had a change to load and render the virtual DOM is able to diff against the BTR" I'm not sure exactly what you mean by this?

docs/en/building/supplemental.md
```diff
@@ -0,0 +1,541 @@
...
+| `host-node`         | Attempts to detect if the environment appears to be a node environment.                                                                                                                                                                                  |
+| `build-time-render` | Statically defined by the build-time rendering system during build-time rendering.                                                                                                                                                                       |
+
+# Build-time Rendering
+
+BTR renders a route to HTML during the build process and in-lines critical CSS and assets needed to display the initial view. This allows Dojo to pre-render the initial HTML used by a route and inject it directly into the page immediately. Skipping the initial render by creating it during the build process gives us many of the same benefits of server side rendering (SSR) such as performance gains and search engine optimization without the complexities of SSR.
```
agubler: We don't "skip" the initial render, the application merges with the existing DOM structure during the initial render.

docs/en/building/supplemental.md
```diff
@@ -0,0 +1,541 @@
...
+{
+	"build-app": {
+		"features": {
+			"production": true
+		}
+	}
+}
+```
+
+The above `production` feature will be set `true` for **production builds** (`dist` mode). The Dojo build system uses the [static-build-loader](https://github.com/dojo/webpack-contrib/#static-build-loader) to rewrite `has()` calls that have been statically asserted by the system as `true` or `false`. This allows Webpack to conditionally identify code as unreachable and remove those dead code branches from the build.
```
agubler: I believe `node_env` is built in and set to `development` on a `dev` build and `production` on a `dist` build

docs/en/building/supplemental.md
```diff
@@ -0,0 +1,541 @@
...
+
+## Assets
+
+Many assets like CSS and images will be imported by modules and inlined automatically by the build process. However, sometimes it is necessary to serve static assets like the favicon or video files.
+
+Static assets can be added to an `assets/` directory at the project root. At build time, these assets are copied to an `assets/` directory along-side the built application.
+
+The build also parses `src/index.html` for CSS, JavaScript, and image assets, hashes them and includes them in the output directory. A favicon can be added to `src` and referenced by `src/index.html`. The build will then hash the file and copy it to the output directory with a file name of `favicon.[hash].ico`.
+
+## Externals
```
agubler: Externals is an advanced feature and should probably go towards the end.

docs/en/building/basic-usage.md
sbinge: Seconded - as part of the overall doc update initiative we're looking to simplify the READMEs down to contributing info & not much more, then linking out to these reference guides as a replacement for the detailed usage content they currently contain.

Specifically for CLI content - we can include relevant documentation for specific commands inline within these ref guides when discussing a particular concept (e.g. creating apps). For 'detailed' CLI docs, i.e. describing every possible argument, we're hoping that the CLI commands themselves can be entirely self-documenting via their help output.

docs/en/building/basic-usage.md
```diff
@@ -0,0 +1,118 @@
...
+		}
+	},
+	"test-intern": {},
+	"create-widget": {
+		"tests": "tests/unit"
+	}
+}
+```
+
+In this example we have a section for each of the three CLI command modules: [@dojo/cli-build-app](https://github.com/dojo/cli-build-app/), [@dojo/cli-test-intern](https://github.com/dojo/cli-test-intern), and [@dojo/cli-create-widget](https://github.com/dojo/cli-create-widget). Configuration is _always_ hierarchical and in the order of command => feature => configuration.
```
sbinge: Minor style consistency point - the new guides are favoring writing from a third-person perspective (if needed at all), essentially avoiding first-/second-person descriptions such as 'we'/'you'/'your'.

docs/en/building/supplemental.md
```diff
@@ -0,0 +1,541 @@
...
+	"build-app": {
+		"bundles": {
+			"fr": ["src/**/nls/fr/**"],
+			"de": ["src/**/nls/de/**"]
+		}
+	}
+}
+```
+
+In this case Dojo will create bundles named `fr.[hash].js` and `de.[hash].js`. For more information see [Working with message bundles](/reference-guides/i18n/working-with-message-bundles) from the Internationalization reference guide.
```
sbinge: We need to work out how linking across guides can function - ideally we can have links that are valid across the markdown files as they are structured within the framework repo, then translate those to appropriate paths when building site. Something to discuss with the site devs.

docs/en/building/basic-usage.md
```diff
@@ -0,0 +1,118 @@
+# Basic Usage
+
+## Creating a Dojo application
+
+The first step for building a Dojo application is to install the CLI
+
+> Command line
```
sbinge: We can probably skip these quotes for CLI commands; they're only really useful when naming source code files.

docs/en/building/supplemental.md
```diff
@@ -0,0 +1,541 @@
...
+	}
+];
+```
+
+> src/App.ts
+
+```ts
+export default class App extends WidgetBase {
+	protected render() {
+		return v('div', { classes: [css.root] }, [
```
sbinge: These new ref guides are effectively launching as v6's documentation, so preferring the latest & greatest authoring styles when giving widget examples. This is mainly:

* Preferring `tsx` output over `v()`/`w()`
* Moving to functional widgets (we'll need to do a pass over all guides once this feature has stabilized, so we can keep examples as class-based for the moment)

docs/en/building/introduction.md
```diff
@@ -0,0 +1,15 @@
...
+| PWA Support                | Progressive web apps create a faster more reliable user experience by caching content and even working offline. Dojo makes it easy to create a service worker through configuration or define it in code and build it as part of your application.                                                              |
+| Build-time rendering       | Renders routes during build time to generate initial HTML and CSS. By rendering at build time Dojo can skip the cost of producing an initial render and create a more responsive application with little additional complexity.                                                                                 |
+| Ejectable configurations   | Take total control of the build process by ejecting the underlying configurations and work directly with the low-level tools. An escape hatch for anyone who needs complex or specific functionality outside of the norm.                                                                                       |
```
sbinge: I'd say we can keep this feature somewhat buried in the supplemental - we'd prefer if users favoured the framework's build pipeline and fed back if it didn't meet their needs. Ejecting is effectively a last resort.

docs/en/building/supplemental.md
```diff
@@ -0,0 +1,541 @@
...
+		}
+	}
+}
+```
+
+Types for any dependencies included in `externals` can be installed in `node_modules/@types`, like any other dependency.
+
+Because these files are external to the main build, no versioning or hashing will be performed on files in a production build, with the exception
+of the links to any `inject`ed assets. The `to` property can be used to specify a versioned directory to copy dependencies to in order to avoid different
+versions of files being cached.
```
sbinge: Minor EOL/hard wrapping inconsistencies in this paragraph.

docs/en/building/supplemental.md
```diff
@@ -0,0 +1,541 @@
...
+		}
+	}
+}
+```
+
+### ServiceWorker Configuration
+
+Under the hood, the `ServicerWorkerPlugin` from `@dojo/webpack-contrib` is used to generate the service worker, and all of its options are valid `pwa.serviceWorker` properties.
+
+| Property       | Type       | Optional | Description                                                                                     |
```
sbinge: We'd previously been using unordered lists with abbreviated-TypeScript syntax to describe API parameter/config option lists ([for example](https://github.com/dojo/framework/blob/master/docs/en/i18n/supplemental.md#i18nmixin-widget-properties)), but I actually prefer this tabular layout; it's cleaner and more structured. We should probably revisit the committed ref guides and update them with this format for consistency (separate task to this ref guide of course!).

docs/en/building/supplemental.md
sbinge: I'd listed it as a bullet point in https://github.com/dojo/framework/issues/322, but yeah if we think it doesn't belong in this guide we can move it elsewhere for a future advanced guide.

docs/en/building/basic-usage.md
sbinge: Another requirement I'd originally listed in https://github.com/dojo/framework/issues/322. It's not strictly 'building' but is somewhat of a precursor requirement to it. Not sure where a better place would be for it - it feels too small to exist on its own. I'd say including it as a single section in this guide makes the most sense?

sbinge: That said, if we have plans to expand app creation to support more variations then it makes sense to separate it out into its own guide.

agubler: Yeah I don't think the building reference guide should contain anything related to creating an application or cli-create-app - the assumption is you already have an application 😄 I do think there could/should be a small reference guide for using the "create" commands

devpaul: This was included so a user can go from absolutely nothing to building a dojo app using only this guide. Also to meet the "quickly scaffolding" requirement in the ticket. As a user there's value in a single guide that takes them through start to finish, even if this isn't the right place.

agubler: I don't doubt there is value in guiding users through the create commands, but I don't think it should be at the beginning of of the build reference guide.

docs/en/building/basic-usage.md
devpaul: This feels a little weird having information for one project in the framework repository, but I understand the decision.

docs/en/building/basic-usage.md
devpaul: I can cut this down. The purpose was to acknowledge that there are unique build targets for functional and unit tests and perhaps provide a little insight on why this is the case.

docs/en/building/basic-usage.md
devpaul: I'm concerned using passive voice in tutorials and guides is less direct and engaging. I understand passive voice in customer contacts as a way to avoid confrontation and blame. Can we talk about this when we get a moment?

docs/en/building/supplemental.md
devpaul: Do we have a style guide somewhere that lists these preferences? This code was pulled directly from `@dojo/cli-create-app` and was used for consistency from the basics guide.

docs/en/building/supplemental.md
devpaul: This was included to address this bullet

> how cross-bundle duplication may occur for shared resources across multiple bundles

Should we just remove the reference to webpack and call it good, or should we say more/other things? Just to repeat my TODO above, I am not very confident in this section.

docs/en/building/supplemental.md
devpaul: I remember having some issues getting this to work. This behavior should be verified and updated.

docs/en/building/supplemental.md
devpaul: I'm not sure where else this would be written down if it's not in this guide. Elided imports are important for targeting various environments.

docs/en/building/supplemental.md
devpaul: I'm trying to say that the initial render of the DOM occurs during build time; during runtime it's appended to the DOM; after the application compares against that rather than generating the DOM from scratch.

Maybe "skip" the initial render isn't the right wording. How would you describe what's happening and why it improves performance?

docs/en/building/supplemental.md
devpaul: I *think* I was trying to say that BTR is built against routes. Definitely needs better wording.

docs/en/building/supplemental.md
devpaul: Maybe it would be better as a note somewhere? I copied this and am not completely aware of the implications TBH.

docs/en/building/supplemental.md
devpaul: I think this came a conversation I had about when the BTR was captured. I remember working with SSR solutions and deciding when to capture the DOM was always an issue.

devpaul:
Thanks for all the feedback! Hope to have time soon to make these changes

docs/en/building/supplemental.md
```diff
@@ -0,0 +1,541 @@
...
+
+-   `outputPath`: An optional property specifying an output path to which files should be copied.
+
+-   `dependencies`: A required array that defines which modules should be loaded via the external loader, and what files should be included in the build. Each entry can be one of two types:
+    _ A string that indicates that this path, and any children of this path, should be loaded via the external loader.
+    _ An object that provides additional configuration for dependencies that need to be copied into the built application. This object has the following properties:
+
+| Property | Type                                                    | optional | Description                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                             |
```
sbinge: We should probably be consistent with how we represent the `Optional` columns - I'd suggest we use Yes/No values, similar to the other tables in this guide.

agubler:
Superseded by #442 which incorporates all of @devpaul's original changes with feedback addressed. Thanks both 😄

Issue #339, opened 2019-05-16
Dojo store support for end of list path


devpaul:
In Dojo Stores there isn't a clear way to refer to the end of a list. While there does appear to be some support for JSON Patch's `-` path element there doesn't seem to be a way to refer to it using `path` or `at` and even when it is used it appears to be buggy.

> Operation

```ts
apply([
  add(path('list'), [ 'one', 'two' ]),
  add(at(path('list'), '-'), 'three')  // type error
]);
```

> Result

```json
[ "one", "three", "two" ]
```

> Operation

```ts
apply([
  add(at(path('list'), '-'), 'three')  // type error
]);
```

> Result

```json
{
  "list": {
    "-": "three"
  }
}
```

## Proposed solutions

One solution would be to add a function that explicitly indexes the end of a list. We could add this function to `at`.

```ts
apply([
  add(at.end(path('list')), 'three')
]);
```

This has the benefit of flowing pretty well, being human readable, and not significantly expanding the interface of `CommandRequest`.

Another solution would be to simply change `at` to accept a `number | '-'` or `number | undefined` and in the instance where it is `undefined` simply add to the end of the list. I'm not fond of either of these solutions as the first doesn't read as well to anyone unfamiliar with JSON Patch and the second makes it easy to omit a value and unintentionally act on the end of the list.

We should also ensure that `add`, `replace`, `remove`, and `test` behave as expected with this addition.

PR #341, opened 2019-05-17
Stores reference

### LINKED ISSUE #324: Reference guide: Dojo Stores
* Describing advanced state management for Dojo apps (basic + intermediate are being covered in 'creating widgets').
* Covering how to use stores; update & streamline content from: https://github.com/dojo/framework/blob/master/src/stores/README.md
* Best practices on data model lifecycle (CRUD, handling changes to multiple objects as atomically as possible) & how each step best translates to Processes
* Handling errors, rolling back
* Pointers for integrating with underlying data stores eg REST services?
* NOTE: likely impacted by the new Resources API in v6/v7 so we need to discuss how to approach this guide.

devpaul:
This is a draft of the stores reference guide.

* Introduction is mostly complete. Some cleanup on the features will likely need to happen when I'm done writing and have full view of everything.
* Basic usage is done. It contains everything a person needs to use state from store to widget written in a manner that progresses from operations thru `StoreProvider`.
* Supplemental is done

Resolves #324 

We added a discussion about Dojo's implementation of JSON patch in the supplemental section after writing #339 and discussing the `state` feature with @maier49. Dojo differs quite significantly from JSON patch in that it tries to do the right thing regarding uninitialized state and avoid throwing errors.

### DIFF
--- a/docs/en/stores/basic-usage.md
+++ b/docs/en/stores/basic-usage.md
@@ -1,0 +1,87 @@
+# Basic Usage
+
+-   Overview of Dojo stores
+    _ https://github.com/dojo/framework/blob/master/src/stores/README.md#overview
+    _ https://github.com/dojo/framework/blob/master/src/stores/README.md#how-does-this-differ-from-redux
+    _ global store
+    _ type safe \* uni-directional data flow
+
+## The Store
+
+-   Overview
+    _ one global store per application
+    _ in general store data should be serializable \* special cases need special care
+
+### Setting up a store
+
+-   Creating and registering a store
+
+## Updating stores
+
+-   Overview
+
+### Operations
+
+-   https://github.com/dojo/framework/blob/master/src/stores/README.md#operations
+-   JSON patch
+    _ add
+    _ remove
+    _ replace
+    _ test
+
+### Commands
+
+-   https://github.com/dojo/framework/blob/master/src/stores/README.md#commands
+-   Overview
+    _ functions with access to the store
+    _ return patch operation(s) \* may be async
+-   CommandRequest (Command API)
+    _ at
+    _ path
+    _ get
+    _ payload \* state object
+-   creating commands
+    _ command factory
+    _ examples
+
+### Processes
+
+-   overview \* https://github.com/dojo/framework/blob/master/src/stores/README.md#processes
+-   Combine commands into a single process
+-   execute commands against a store
+-   attach a middleware
+    _ before commands
+    _ useful for gating
+    _ after commands
+    _ useful for error handling
+
+### Initial State
+
+Example for setting up an application's initial state
+https://github.com/dojo/framework/blob/master/src/stores/README.md#initial-state
+
+## Widgets and Stores
+
+### Container
+
+-   Overview
+    _ wraps a Widget
+    _ has access to a store
+    _ provides dependencies
+    _ interfaces matches the widget
+-   When should I use
+    _ Containers are useful when a 1:1 mapping is desired
+    _ Useful for connecting processes + store data
+
+### StoreProvider
+
+-   Overview
+    _ has access to a store
+    _ has its own renderer
+    _ is usually encapsulated in a widget
+    _ therefore can have a different interface
+-   When should I use
+    _ StoreProviders are more flexible than Containers
+    _ Useful for encapsulating widgets
+    _ Can do the same thing as Containers
+    _ Can change the interface of a wrapped widget
--- a/docs/en/stores/basic-usage.md
+++ b/docs/en/stores/basic-usage.md
@@ -48,11 +48,11 @@ The above is a simple example for illustrative purposes. Examples in this refere
 
 ## Updating stores
 
-To work with Dojo Stores there are three core but simple concepts - Operations, Commands, and Processes.
+To work with Dojo Stores there are three core concepts.
 
 -   **Operations** are instructions to manipulate the state held by the store
 -   **Commands** are simple functions that perform business logic and return operations
--   **Processes** execute a group of commands and usually represent application behavior
+-   **Processes** execute a group of commands and represent application behavior
 
 ### Operations
 
@@ -60,11 +60,11 @@ Stores cannot be modified directly. Instead four operations are available for mo
 
 #### `path` and `at`
 
-The path is a `string` that describes the location where an operation is applied. The `path` and `at` functions are part of the `Store` and are accessible inside of a `Command`.
+The path is a `string` that describes the location where an operation is applied. The `path` and `at` functions are part of the `Store` and are accessible inside of a `Command` (which we will talk about later).
 
-Here are a couple examples to show how `path` and `at` describe a location in the store. The `State` provided is there to illustrate the location of a value and describe how the location functions are used to reference that location. Typically `path` and `at` are used in conjunction with operations inside of Commands, which we will show later.
+Here are a couple examples to show how `path` and `at` describe a location in the store. The `State` is the same as defined above in `interface.d.ts`. It is used by the `Store` to understand the shape of the state data.
 
-> `path` transverses an object hierarchy
+> `path` of the current user name
 
 ```ts
 const store = new Store<State>();
@@ -172,7 +172,7 @@ These operations set the current user unless a current user already exists.
 
 ### Commands
 
-Commands are simply functions that are called while executing a process and return an array of operations that modify the store. Each command is passed a `CommandRequest` which provides `path` and `at` functions to generate `Path`s in a type-safe way, a `get` function for access to the store's state, a `payload` object for the argument that the process executor was called with.
+Commands are simply functions that are called while executing a process. They return an array of operations that modify the store. Each command is passed a `CommandRequest` which provides `path` and `at` functions to generate `Path`s in a type-safe way, a `get` function for access to the store's state, a `payload` object for the argument that the process executor was called with.
 
 #### The command factory
 
@@ -193,7 +193,7 @@ const myCommand = createCommand(({ at, get, path, payload, state }) => {
 
 `createCommand` will ensure that the wrapped command has the correct typing and the passed `CommandRequest` functions will be typed to the `State` interface provided to `createCommandFactory`. While it is possible to manually type commands this guide will use `createCommand` in its examples.
 
-#### `get'
+#### `get`
 
 The `get` function returns a value from the store at a specified path or `undefined` if a value does not exist at that location.
 
@@ -362,7 +362,7 @@ login(store)({
 
 #### Initial State
 
-Initial application state can be defined on store creating by executing a process.
+Initial application state can be defined on a store creating by executing a process.
 
 > main.ts
 
--- a/docs/en/stores/basic-usage.md
+++ b/docs/en/stores/basic-usage.md
@@ -1,64 +1,407 @@
 # Basic Usage
 
--   Overview of Dojo stores
-    _ https://github.com/dojo/framework/blob/master/src/stores/README.md#overview
-    _ https://github.com/dojo/framework/blob/master/src/stores/README.md#how-does-this-differ-from-redux
-    _ global store
-    _ type safe \* uni-directional data flow
+Dojo is a reactive architecture concerned with constantly modifying and rendering the current state of an application. In simple systems this can happen at a local level and widgets can maintain their own state. However, as a system becomes more complex the need to better compartmentalize and encapsulate data and create a clean separation of concerns quickly grows.
+
+Stores provide a clean interface for storing, modifying, and retrieving data from a global object using a unidirectional data flow. Common patterns such as asynchronous retrieval, middleware, and undo patterns are built in. This allows widgets to remain focused on their primary roles of providing a visual representation of themselves and listening for user interactions.
 
 ## The Store
 
--   Overview
-    _ one global store per application
-    _ in general store data should be serializable \* special cases need special care
+The store hold the global, atomic state for the entire application. It is created when the application is created and defined in the `Registry` with an injector.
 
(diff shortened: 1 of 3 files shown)


docs/en/stores/basic-usage.md
```diff
@@ -0,0 +1,471 @@
...
+
+const registry = new Registry();
+const store = new Store<State>();
+registry.defineInjector('state', (invalidator) => {
+	store.on('invalidate', () => invalidator());
+	return () => store;
+});
+```
+
+`State` defines the shape of the global store using an interface. In general state should be serializable. This improves performance and makes it easier for Dojo to determine when changes to the data occur.
```
agubler: `State` should be completely serializable, this is important for features such as history.

docs/en/stores/basic-usage.md
```diff
@@ -0,0 +1,471 @@
...
+
+Stores provide a clean interface for storing, modifying, and retrieving data from a global object using a unidirectional data flow. Common patterns such as asynchronous retrieval, middleware, and undo patterns are built in. This allows widgets to remain focused on their primary roles of providing a visual representation of themselves and listening for user interactions.
+
+## The Store
+
+The store hold the global, atomic state for the entire application. It is created when the application is created and defined in the `Registry` with an injector.
+
+> main.ts
+
+```ts
```
agubler: Do you think that we should use the helper from `StoreInjector`, `registerStoreInjector` for the example:

```ts
import { registerStoreInjector } from '@dojo/framework/store/StoreInjector';
import Store from '@dojo/framework/stores/Store';
import { State } from './interfaces';

const store = new Store<State>();
const registry = registerStoreInjector(store);
```

docs/en/stores/basic-usage.md
```diff
@@ -0,0 +1,471 @@
...
+
+To work with Dojo Stores there are three core concepts.
+
+-   **Operations** are instructions to manipulate the state held by the store
+-   **Commands** are simple functions that perform business logic and return operations
+-   **Processes** execute a group of commands and represent application behavior
+
+### Operations
+
+Stores cannot be modified directly. Instead four operations are available for modifying state: `add`, `remove`, `replace`, and `test`. These operations are represented by simple objects based on the [JSON patch](http://jsonpatch.com/) format, but in practice helper functions will always be used.
```
agubler: I don't think we need to mention JSON patch here, although being inspired by JSON patch is where we started it is an implementation detail that we could change in the future.

docs/en/stores/basic-usage.md
```diff
@@ -0,0 +1,471 @@
...
+
+const login = createProcess('login', [ fetchUser, loadUserData ], [ resetUserOnError ]);
+```
+
+#### Excuting a Process
+
+A process simply defines an execution flow for a set of data. To execute a process it needs access to the store to create an executor. Then the executor can be provided a payload to act upon.
+
+```ts
+import { login } from 'commands/login';
```
agubler: Should this be `processes`? And also prefixed with `./`?

docs/en/stores/basic-usage.md
```diff
@@ -0,0 +1,471 @@
...
+process(store)({ one: 'one', two: 'two' });
+```
+
+## Widgets and Stores
+
+There are two state containers available for widgets: `Container` and `StoreProvider`. These containers connect the application store with a widget.
+
+Note that the documentation in this section is intended to show how widgets and state are connected. For a more detailed look see the [Widgets]() reference guide.
+
+### Container
```
agubler: Can we talk about the `StoreProvider` first please?

docs/en/stores/basic-usage.md
```diff
@@ -0,0 +1,471 @@
...
+				])
+			}
+		}
+	};
+};
+
+const login = createProcess('login', [ fetchUser, loadUserData ], [ resetUserOnError ]);
+```
+
+#### Excuting a Process
```
agubler: Perhaps it's worth talking about this in the `StoreProvider` / `Container` sections as that's the main place that you will have access to the store? I'm not sure we should be suggesting people to export the store instance from `main.ts`.... Maybe we should have a section on creating initial store state, which would normally be done in the `main.ts`

docs/en/stores/basic-usage.md
```diff
@@ -0,0 +1,471 @@
...
+> widget/User.provider.ts
+
+```ts
+import WidgetBase from '@dojo/framework/widget-core/WidgetBase';
+import { w } from '@dojo/framework/widget-core/d';
+import StoreProvider from '@dojo/framework/stores/StoreProvider';
+import { State } from '../../interfaces';
+import User from './User';
+
+export class UserProvider extends WidgetBase {
```
agubler: Is it necessary to have a `UserProvider` instead of using the `StoreProvider` inline in which ever parent widget requires it? It will be easier to demonstrate leveraging values from the parent widget, I don't want people to jump to creating a separate widget for each provider.

docs/en/stores/basic-usage.md
```diff
@@ -0,0 +1,471 @@
+# Basic Usage
+
+Dojo is a reactive architecture concerned with constantly modifying and rendering the current state of an application. In simple systems this can happen at a local level and widgets can maintain their own state. However, as a system becomes more complex the need to better compartmentalize and encapsulate data and create a clean separation of concerns quickly grows.
```
agubler: Do you think this section should be in the introduction?

docs/en/stores/basic-usage.md
```diff
@@ -0,0 +1,471 @@
...
+
+-   **Operations** are instructions to manipulate the state held by the store
+-   **Commands** are simple functions that perform business logic and return operations
+-   **Processes** execute a group of commands and represent application behavior
+
+### Operations
+
+Stores cannot be modified directly. Instead four operations are available for modifying state: `add`, `remove`, `replace`, and `test`. These operations are represented by simple objects based on the [JSON patch](http://jsonpatch.com/) format, but in practice helper functions will always be used.
+
+#### `path` and `at`
```
agubler: For basic usage do you think it would be easier to talk about arrays as a separate concern (in a dedicated section)?

docs/en/stores/basic-usage.md
```diff
@@ -0,0 +1,471 @@
...
+> `add` example
+
+```ts
+import { add } from '@dojo/framework/stores/state/operations';
+
+const store = new Store<State>();
+const { apply, at, path } = store;
+const user = { id: '0', name: 'Paul' };
+
+apply([add(at(path('users', 'list'), 0), user)]);
```
agubler: I don't think we should be demonstrating interacting with the store directly like this. It's not an encouraged pattern

docs/en/stores/basic-usage.md
```diff
@@ -0,0 +1,471 @@
...
+
+## Updating stores
+
+To work with Dojo Stores there are three core concepts.
+
+-   **Operations** are instructions to manipulate the state held by the store
+-   **Commands** are simple functions that perform business logic and return operations
+-   **Processes** execute a group of commands and represent application behavior
+
+### Operations
```
agubler: For this section might be easier to group commands and operations together, that way you will be able to show examples with the recommended approach (i.e. using commands to return operations)

docs/en/stores/basic-usage.md
```diff
@@ -0,0 +1,471 @@
...
+		current: user[0],
+		list: users
+	}),
+	replace(at(path('users', 'list'), 1), newUser)
+]);
+```
+
+This example will replace the second user in the `list` with `newUser`.
+
+#### `test` operation
```
agubler: I don't think the `test` operation really belongs in basic usage, generally users won't need to explicitly use the `test` operation.

docs/en/stores/basic-usage.md
```diff
@@ -0,0 +1,471 @@
...
+const createCommand = createCommandFactory<State>();
+
+const addUser = createCommand<User>(({ at, path, payload }) => {
+	return [add(at(path('users', 'list'), 0), payload)];
+});
+```
+
+This example adds a user provided in the `payload` to the beginning of `/users/list`.
+
+#### `state`
```
agubler: Do you think this should be in supplemental rather than basic usage, given the restrictions using it, i.e. no support for IE11.

docs/en/stores/basic-usage.md
```diff
@@ -0,0 +1,471 @@
...
+				return w(User, { name });
+			}
+		});
+	}
+}
+```
+
+`UserProvider` extends `WidgetBase` allowing us the opportunity to redefine the widget's properties. In this case `UserProvider` does not accept any properties. It gets all of the information it needs from the store. The `StoreProvider` occurs as part of `UserProvider`'s render and provides its own renderer just like any other widget.
+
+The `StoreProvider` is more robust than a `Container`. Since it is always wrapped by a widget it gives the opportunity to decide what properties should be provided by the outside widget and which should be provided from the store.
```
agubler: I don't think this is true, it's not always wrapped by a widget nor would I recommend doing that by default. I would also start with using the `StoreProvider` inline and then think about where to abstract when it becomes required.

docs/en/stores/supplemental.md
```diff
@@ -0,0 +1,394 @@
+# StoreProvider
+
+## StoreProvider API
+
+The StoreProvider accepts three properties
+
+-   `renderer`: A render function that has the store injected in order to access state and pass processes to child widgets.
+-   `stateKey`: The key of the state in the registry.
+-   `paths` (optional): A function to connect the Container to sections of the state.
```
agubler: "connect the provider to sections of the state"

docs/en/stores/supplemental.md
```diff
@@ -0,0 +1,394 @@
...
+
+## Invalidation
+
+The `StoreProvider` has two main ways to trigger invalidation and cause a rerender.
+
+1.  The recommended approach is to register `path`s on container creation to ensure invalidation will only occur when state you are interested in changes.
```
agubler: "register `path`s by passing the `paths` property to the provider...."

docs/en/stores/supplemental.md
```diff
@@ -0,0 +1,394 @@
...
+
+> TypedStoreProvider.ts
+
+```ts
+import { State } from './interface';
+
+export class TypedStoreProvider extends StoreProvider<State> {}
+```
+
+**However** in order for TypeScript to infer this correctly when using w(), the generic will need to be explicitly passed.
```
agubler: Using `tsx` infers the renderer functions arguments correctly, do you think we should demonstrate this?

docs/en/stores/supplemental.md
```diff
@@ -0,0 +1,394 @@
...
+
+> main.ts
+
+```ts
+store.on('invalidate', () => {
+	// do something when the store's state has been updated.
+});
+```
+
+# JSON Patch and Store operations
```
agubler: As before I don't think we should mention JSON path in the reference guide.

agubler:
@devpaul Sorry to be a pain but as https://github.com/dojo/framework/pull/357 has been merged, could you update any references to `framework/widget-core` to `framework/core`. Thanks!

docs/en/stores/basic-usage.md
devpaul: I agree that it usually should be. How would you recommend handling state that contains `TypedArray`s for large data sets or `BigInt`?

docs/en/stores/basic-usage.md
devpaul: Totally. I forgot about that :)

docs/en/stores/basic-usage.md
devpaul: I agree. After writing this I spoke w/ @maier49 and found that we deviate from JSON patch in a lot of small ways

docs/en/stores/basic-usage.md
devpaul: I think that's a good idea. If I introduce commands and operations together I won't need to use `store.apply` to show how operations work.

docs/en/stores/basic-usage.md
devpaul: I think this depends on how soon `state` will become the preferred way of interacting with the store.

docs/en/stores/basic-usage.md
devpaul: You make a good point. I wanted to show the most basic usage of a process, but in doing so it could be reinforcing a bad pattern.

docs/en/stores/basic-usage.md
devpaul: Yeah. This usage pattern was a carry-over from another project that was migrating from containers. I wish I could go on a rant about when someone should create clean widgets vs store-coupled widgets.

docs/en/stores/basic-usage.md
devpaul: Hmm maybe this part can be removed entirely. For me it was novel when the `StoreProvider` became available because `Container` didn't allow properties to be redefined it was actually a lot of work to connect a store to a widget and then change the properties. So while I was really excited about `StoreProvider` for this reason future Dojo users can just take this for granted 😅

docs/en/stores/supplemental.md
devpaul: That sounds great. Why does it work for `tsx` and not `w()`?

docs/en/stores/supplemental.md
devpaul: There's some advanced functionality here that I want to cover

1. that stores will automatically create the underlying structure
1. that stores will rarely throw an error
1. that adding and replacing past the end of an array will add to the end of an array

Maybe keep this section and remove the references to JSON Patch

devpaul:
Thanks for your feedback @agubler! Hope to have time to work on this soon

PR #345, opened 2019-05-20
Fix length return for immutable lists

### LINKED ISSUE #343: `ImmutableState` not returning correct values for array length
**Bug** 

Package Version: 6.0.0-alpha.1

The Immutable lists internally use `size` instead of `length`, and the ImmutableState implementation is not currently handling that to return the correct value.

maier49:
**Type:** bug

**Description:**
Adds logic to convert Immutable List size to array length 
Resolves #343

### DIFF
--- a/tests/stores/unit/state/ImmutableState.ts
+++ b/tests/stores/unit/state/ImmutableState.ts
@@ -218,11 +218,11 @@ describe('state/ImmutableState', () => {
 		const state = new ImmutableState();
 		state.apply([{ op: OperationType.ADD, path: new Pointer('foo'), value: [] }]);
 
-		assert.equal(state.path('foo', 'length').value, 0);
+		assert.equal(state.get(state.path('foo', 'length')), 0);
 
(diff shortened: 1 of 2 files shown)


tests/stores/unit/state/ImmutableState.ts
```diff
@@ -214,6 +214,17 @@ describe('state/ImmutableState', () => {
 		});
 	});
 
+	it('should report the length of an array', () => {
+		const state = new ImmutableState();
+		state.apply([{ op: OperationType.ADD, path: new Pointer('foo'), value: [] }]);
+
+		assert.equal(state.path('foo', 'length').value, 0);
```
agubler: Do you think we should use `state.get` to actually return the value?

maier49: Fair enough, I've updated this. Though it is a bit funny to replace `path.value` with a call to `get` that's just returning `path.value` anyways. Maybe at some point we should evaluate whether `state.get` will ever have merit a more useful implementation and if not, consider removing it?

PR #348, opened 2019-05-20
Fixing a problem with auto-bind and multiple subclasses

rorticus:
**Type:** bug

**Description:**

There is a problem with the new auto-bind code when there are multiple levels of inheritance. It currently only applies the binding logic to the first level of inheritance.

To fix this, we're walking the prototype chain up to `WidgetBase` and gathering all the properties from there. Since this is a bit of an expensive operation, we're also caching the result of the property gathering so it only needs to happen once per widget class.

### DIFF
--- a/src/widget-core/WidgetBase.ts
+++ b/src/widget-core/WidgetBase.ts
@@ -52,15 +52,38 @@ function isDomMeta(meta: any): meta is Base {
 	return Boolean(meta.afterRender);
 }
 
-const IGNORE_LIST: (string | symbol)[] = ['constructor', 'render'];
+const IGNORE_LIST: (string | symbol)[] = ['render', ...Object.getOwnPropertyNames(Object.getPrototypeOf({}))];
+
+const autoBindCache = new Map<any, string[]>();
 
 function autoBind(instance: any) {
-	let keys: string[] = Object.getOwnPropertyNames(instance.constructor.prototype);
+	let prototype = instance.constructor.prototype;
+
+	let keys: string[] = [];
+
+	if (autoBindCache.has(prototype)) {
+		keys = autoBindCache.get(prototype) as string[];
+	} else {
+		let p = prototype;
+		while (p) {
+			const ownKeys = Object.getOwnPropertyNames(p);
+
+			if (ownKeys.indexOf('_type') !== -1) {
+				break;
+			}
+
+			keys = [...keys, ...ownKeys];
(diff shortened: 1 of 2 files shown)


src/widget-core/WidgetBase.ts
```diff
@@ -52,15 +52,38 @@ function isDomMeta(meta: any): meta is Base {
 	return Boolean(meta.afterRender);
 }
 
-const IGNORE_LIST: (string | symbol)[] = ['constructor', 'render'];
+const IGNORE_LIST: (string | symbol)[] = ['render', ...Object.getOwnPropertyNames(Object.getPrototypeOf({}))];
```
rorticus: We can exclude all the properties that come from `Object.prototype`. Technically we might not even need this since we break our loop on subclasses of `WidgetBase`?

src/widget-core/WidgetBase.ts
```diff
@@ -52,15 +52,38 @@ function isDomMeta(meta: any): meta is Base {
...
+
+			if (ownKeys.indexOf('_type') !== -1) {
+				break;
+			}
+
+			keys = [...keys, ...ownKeys];
+			p = Object.getPrototypeOf(p);
+		}
+
+		keys = keys.filter((k) => typeof instance[k] === 'function' && IGNORE_LIST.indexOf(k) === -1);
```
rorticus: The filtering of keys through the ignore list is now done up here rather than down yonder.

src/widget-core/WidgetBase.ts
```diff
@@ -52,15 +52,38 @@ function isDomMeta(meta: any): meta is Base {
...
 function autoBind(instance: any) {
-	let keys: string[] = Object.getOwnPropertyNames(instance.constructor.prototype);
+	let prototype = instance.constructor.prototype;
+
+	let keys: string[] = [];
+
+	if (autoBindCache.has(prototype)) {
+		keys = autoBindCache.get(prototype) as string[];
+	} else {
+		let p = prototype;
```
agubler: Don't need to do this, just use `prototype`?

agubler: @rorticus Did you see this? ^

agubler:
@rorticus It also looks like CI is failing

src/widget-core/WidgetBase.ts
rorticus: Removed extra variable `p`!

Issue #349, opened 2019-05-21
Proposal for functional Widgets and Render Middleware


matt-gadd:
# Proposal for functional Widgets and Render Middleware

# Introduction

As part of 6.0 we want to offer a simpler way of constructing widgets. This would be moving away from Classes over to a factory and render function. This allows us to really make the user focus on writing simple reactive Widgets, and not using things like Class constructors or WidgetBase lifecycles in exotic ways. Having a render function imposes some limits on the existing extensibility patterns we currently use:

# Existing Widget extensibility patterns

Since launching modern dojo, we've had a number of patterns that provide extensibility and funtionality to widgets that include the following:

* Decorators - we have a small amount of decorators in widget-core for attaching to some lifecycle events. We have never comprehensively committed to decorators due to limitations on how they can affect types and that they can only be used on classes.

* Mixins - the main reason we leverage mixins for some behaviour is the ability for them to augment a users widget properties. They also have access to the same lifecycle as the widget would have and can provide utility functions off of the class. In terms of authoring they are not particularly nice to type and you cannot compose mixins inside other mixins.

* Meta - meta is currently a specific pattern used for making imperative dom api's reactive. The meta's are both flexible in terms of the api they can provide, and type well. They're used inside a render function making them logical and not dependent on other lifecycles. Meta's are currently not composable and have a very limited dom lifecycle api available.

# Render Middleware

When supporting widgets without classes, both Decorators and Mixins are no longer applicable. Which leaves us with the Meta pattern. The Meta pattern can be expanded to be more generic beyond dom api's, and with later TypeScript versions it's also possible to influence the Widget Properties like the Mixin Pattern. This means we should be able to convert the existing Mixins to a Meta like implementation.

Given the new power of the Meta like pattern, we are proposing a more generic name of "Render Middleware" (this is open to suggestions though). As well as now being able to influence Widget Properties, this new pattern will be composable ie being able to use middlewares inside of middlewares, something the current Meta pattern does not support.

On top of this the Middleware aligns nicely/symmetrically with functional Widgets, in that they both use the same creation factory, and both receive the same arguments to the user implemented function. The only difference being that functional Widgets return a RenderResult (DNodes etc), and Middleware returns a user defined API. This alignment on patterns across the board should make it simple for us to explain generally.

# Usage

The new functional Widget and Render Middleware API would look like so:

Example Widget (Button):
```tsx
import cache from './middleware/cache';

const render = create({ cache }).properties<ButtonProperties>();
[...]

PR #352, opened 2019-05-22
Functional Widgets and Render Middleware

agubler:
**Type:** feature

**Description:**

The basic implementation to support functional widgets and render middleware. Includes one core middleware `invalidator` (open to renaming suggestions).

Provides a single function `create` that returns a factory that can be used to create Dojo widgets and render middleware. 

**Note:** Currently `create` is in the `tsx` module in `widget-core`, this is going to change as the feature is built on.

##### Creates a simple widget with a properties interface:

```ts
const { create } from '@dojo/framework/widget-core/tsx';

const render = create().properties<{ foo: string }>();
const MyWidget = render(() => v('div'));
```

##### Creates a render middleware with a properties interface:

```ts
const { create } from '@dojo/framework/widget-core/tsx';

const middleware = create().properties<{ bar: string }>();
const myMiddleware = middleware(() => {
    return {
        get() {},
        set() {}
    };
});
```

##### Creates a widget with middleware that influences the property typings:

```ts
const { create } from '@dojo/framework/widget-core/tsx';

import myMiddleware from './myMiddleware';

const render = create({ myMiddleware }).properties<{ foo: string }>();
const MyWidget = render(({ properties, middleware }) => {
    // can access the injected middleware
    middleware.myMiddleware.get();
    middleware.myMiddleware.set();
    // properties API defined by the widget and the injected middleware
    return v('div', [properties.foo, properties.bar]);
});
```

The change includes support for calling a widget directly, calling the widget directly has the added advantages to be able to infer complex properties such as a render property, but is also backward compatible for use with `w()`.

**Example Usages:**

```tsx
// Supports the existing w() pragma
w(MyWidget, { foo: 'bar' });

// Support being called directly
MyWidget({ foo: 'bar' });

// Supports tsx
<MyWidget foo="bar" />
```

Related to #349


src/widget-core/Registry.ts
```diff
@@ -87,16 +95,23 @@ export interface RegistryInterface {
...
+	return isWidgetBaseConstructor(item) || isWidgetFunction(item);
+}
+
 export function isWidgetConstructorDefaultExport<T>(item: any): item is ESMDefaultWidgetBase<T> {
 	return Boolean(
-		item &&
-			item.hasOwnProperty('__esModule') &&
-			item.hasOwnProperty('default') &&
-			isWidgetBaseConstructor(item.default)
+		item && item.hasOwnProperty('__esModule') && item.hasOwnProperty('default') && isWidget(item.default)
```
matt-gadd: I think for code splitting this needs to also check whether it's a widget factory.

src/widget-core/middleware/core.ts
```diff
@@ -0,0 +1,11 @@
+import { create } from '../tsx';
+import { getInvalidator } from '../vdom';
+
+const createFactory = create();
+
+export const invalidator = createFactory(({ id }) => {
```
matt-gadd: starting to wonder whether we should put these base ones in vdom itself? then we don't have to export the function itself?

src/widget-core/tsx.ts
```diff
@@ -54,3 +64,51 @@ export function tsx(tag: any, properties = {}, ...children: any[]): DNode {
...
+		};
+	};
+	factory.isFactory = true;
+	return factory;
+}
+
+export function create<T extends MiddlewareMap<any>, MiddlewareProps = ReturnType<T[keyof T]>['properties']>(
+	middlewares: T = {} as T
+) {
+	function properties<Props extends {}>() {
```
sbinge: Would this read better as something similar to `withProperties()`? Thinking about it from a perspective of making the create API more fluent, e.g.:

```tsx
create().withProperties<{]>();
```

src/widget-core/middleware/core.ts
agubler: I think this is a good idea, will have more control in vdom itself also.

src/widget-core/tsx.ts
agubler: I was trying to reduce the number of characters but when with `properties`, I don't actually mind the `with` prefix tho for it's fluency. @matt-gadd any thoughts?

agubler:
@sbinge we do have an implementation of a js-framework-benchmark that I have been upgrading and running using functional widgets. The performance is similar and the memory usage slightly less. I will do a side by side comparison of the two different implementation and post onto this issue.

src/widget-core/Registry.ts
agubler: Updated

src/widget-core/middleware/core.ts
agubler: Done

agubler:
@sbinge running two implementations of the js-framework-benchmark, one with classes, one with functional widgets but otherwise identical corroborates the performance these characteristics.

Interactive_Results

src/widget-core/tsx.ts
sbinge: There's another subtle comprehension hiccup here - this method is describing the property interface as part of the factory creation, not widget creation nor rendering invocation.

Reading it as it currently stands, a user could potentially confuse it with passing actual property values as part of a 'create a widget` call; something that `withProperties` or `hasProperties` doesn't help with to resolve.

What about something like `withAPI()`?

agubler: My issue with that is that in the future, when typescript generics catch up we’d like to explore the possibility of using the function to define default properties which will leave the external property API with optional properties but enable referencing them from within widget knowing that although the external API is optional the internal API is not. For the moment typescript would prevent a user passing anything to the function as there are no arguments current.

Additionally the API of a widget is not just properties, it’s also children which again in the future is something that we’d consider exposing for the user to define perhaps with a further `.children` API.

sbinge: Makes sense.

PR #362, opened 2019-05-31
`destroy` and `getDom` Middleware

agubler:
**Type:** feature

**Description:**

Adds two new core middleware implementations, `destroy` and `getDom`

#### `destroy`

Registers a destroy handle that is called when a widget is removed. 

```ts
const factory = create({ destroy });

const MyWidget = factory(({ middleware: { destroy } }) => {
	destroy(() => {
		// do something when the widget is removed
	});
	return v('div', { key: 'div' });
});
```

Limited to registering a single handle per unique middleware, for use-cases that require conditional or multiple handles through out the lifetime of a widget the user is required to manage this within their own middleware.

```ts
const factory = create({ destroy });

const myMiddleware = factory(({ middleware: { destroy } }) => {
	const handles: Handle[] = [];
	destroy(() => {
		let handle: Handle | undefined;
		while (handle = handles.pop()) {
			handle.destroy();
		}
	});
	return {
		foo() {
			// add handle to `handles` array
		}
	}
});
```

#### `getDom`

Returns the DOM node for the `key` of a `VNode` in a widget.

```ts
const factory = create({ getDom });

const myMiddleware = factory(({ middleware: { getDom } }) => {
	return {
		doSomething(key: string | number) {
			const node = getDom(key);
			// do something with the node, like return dimensions.
	};
});
```

Related to #352


agubler:
@matt-gadd updated from `getDom` to `node` for the middleware that returns the DOM node for a particular key.

PR #375, opened 2019-06-05
Issue #302: core widget reference guide

### LINKED ISSUE #302: Reference guide: core widget authoring
As part of our new content initiative, write a reference guide for core widget authoring following the established reference guide format.

Note that the current idea is for state management and custom elements to be separate reference guides, but this could evolve depending on the amount of content in each of the three guides (incl. the one described here).

Existing widget-creation-related content to be consolidated & expanded upon:
* https://github.com/dojo/framework/blob/master/src/widget-core/README.md
* https://github.com/dojo/dojo.io/blob/master/site/source/tutorials/002_creating_an_application/index.md#widgets
* https://github.com/dojo/dojo.io/blob/master/site/source/tutorials/003_creating_widgets/index.md
* https://github.com/dojo/dojo.io/blob/master/site/source/tutorials/004_user_interactions/index.md
* https://github.com/dojo/dojo.io/blob/master/site/source/tutorials/005_form_widgets/index.md
* https://github.com/dojo/dojo.io/blob/master/site/source/tutorials/1015_form_validation/index.md
* https://github.com/dojo/cli-create-widget/blob/master/README.md

sbinge:
**Type:** documentation

**Description:**

Reference guide for basic widget concepts, including structure, VDOM, rendering, state management, interactivity/focus management and development best practices.

Resolves #302


docs/en/creating-widgets/supplemental.md
```diff
@@ -0,0 +1,869 @@
...
+-   `WNode`s, or _Widget Nodes_, which tie Dojo widgets into the VDOM hierarchy.
+
+Both `VNode`s and `WNode`s are considered subtypes of `DNode`s within Dojo's virtual DOM, but applications don't typically deal with `DNode`s in their abstract sense. Using [TSX syntax](#tsx-support) is also preferred as it allows applications to render both virtual node types with a uniform syntax.
+
+### Instantiating VDOM nodes
+
+If [TSX output](#tsx-support) is not desired, widgets can import one or both of the `v()` and `w()` primitives provided by the `@dojo/framework/core/vdom` module. These create `VNode`s and `WNode`s, respectively, and can be used as part of the return value from a [widget's render function](#basic-widget-structure). Their signatures, in abstract terms, are:
+
+-   `v(tagName | VNode, properties?, children?)`:
+-   `w(Widget | constructor, properties?, children?)`
```
agubler: The `properties` param is not optional here.

docs/en/creating-widgets/supplemental.md
```diff
@@ -0,0 +1,869 @@
...
+> src/widgets/MyEncapsulatedStateWidget.tsx
+
+**Function-based variant:**
+
+```tsx
+import { create, tsx } from '@dojo/framework/core/vdom';
+
+const factory = create();
+
+let myState: string = 'Hello from a stateful widget!';
```
agubler: @sbinge We can't do this, this would be shared across however many instances of the `MyEncapsulatedStateWidget` existed in the application causing unexpected results.

The most basic form for storing state across renders using function-based widgets is using the `cache` middleware, however it looks like the intent of this example is to re-render when the button is clicked. In this case it would be perfect for the invalidating cache (`icache`) middleware.

docs/en/creating-widgets/supplemental.md
```diff
@@ -0,0 +1,869 @@
...
+	private counter: number = 0;
+
+	protected render() {
+		return (
+			<div>
+				Current widget state: {this.myState}
+				<br />
+				<button
+					onclick={() => {
+						this.myState = 'State change iteration #' + ++this.counter;
```
agubler: @sbinge My guess here is you want to invalidate when the button is clicked?

docs/en/creating-widgets/supplemental.md
```diff
@@ -0,0 +1,869 @@
...
+
+**Class-based variant:**
+
+```tsx
+import WidgetBase from '@dojo/framework/core/WidgetBase';
+import { tsx } from '@dojo/framework/core/vdom';
+
+export default class MyEncapsulatedStateWidget extends WidgetBase {
+	private myState: string = 'Hello from a stateful widget!';
+	private counter: number = 0;
```
agubler: In later TypeScript you don't need to type variables that are being set to primitives like `string` and `number`

docs/en/creating-widgets/supplemental.md
```diff
@@ -0,0 +1,869 @@
...
+				>
+					Change State
+				</button>
+			</div>
+		);
+	}
+}
+```
+
+Note that this example is not complete - clicking on the 'Change State' button in the running application will not have any effect on the widget's render output. This is because the state is fully encapsulated within `MyEncapsulatedStateWidget`, and Dojo [is not aware of any changes made to it](#widget-properties). Only the widget's initial render will be processed by the framework.
```
agubler: @sbinge Ah okay ignore my comment about invalidating, if you are progressively enhancing this example - but you could move from `cache` to `icache` or just use `cache` and `invalidator` if the intent is to explicitly demo the `invalidator` middleware.

docs/en/creating-widgets/supplemental.md
```diff
@@ -0,0 +1,869 @@
...
+}
+```
+
+## Intermediate: passing widget properties
+
+Passing state into a widget via virtual node [`properties`](#node-properties) is the most effective way of wiring up reactive data flows within a Dojo application.
+
+Widgets specify their own properties [interface](https://www.typescriptlang.org/docs/handbook/interfaces.html) which can include any fields the widget wants to publicly advertise to consumers, including configuration options, fields representing injectable state, as well as any event handler functions.
+
+Function-based widgets pass their properties interface as a generic type argument to the `create().properties()` call. The factory returned from this call chain then makes property values available via a `properties` argument in the render function definition.
```
agubler: `create().properties()` -> `create().properties<MyPropertiesInterface>()` maybe?

docs/en/creating-widgets/supplemental.md
```diff
@@ -0,0 +1,869 @@
...
+				Change State
+			</button>
+		</div>
+	);
+});
+```
+
+**Class-based variant:**
+
+Here, both `myState` and `counter` are updated as part of the same application logic operation, so `@watch()` could be added to either or both of the fields, with the same net effect and performance profile in all cases:
```
agubler: Do you think we should demonstrate using `this.invalidate()` but also talk about the option of using `@watch()`?

docs/en/creating-widgets/supplemental.md
```diff
@@ -0,0 +1,869 @@
...
+```tsx
+import WidgetBase from '@dojo/framework/core/WidgetBase';
+import { tsx } from '@dojo/framework/core/vdom';
+
+export interface MyWidgetProperties {
+	name: string;
+	onNameChange?(newName: string): void;
+}
+
+export default class MyWidget extends WidgetBase<MyWidgetProperties> {
```
agubler: @sbinge I think this should use widget state for the input name and set the input value on each render, something like this:

```tsx
export default class MyWidget extends WidgetBase<MyWidgetProperties> {
	private newName = '';
	protected render() {
		const { name, onNameChange } = this.properties;
		return (
			<div>
				<span>Hello, {name}! Not you? Set your name:</span>
				<input
					type="text"
					value={this.newName}
					oninput={(e: Event) => {
						this.newName = (e.target as HTMLInputElement).value;
						this.invalidate();
					}}
				/>
				<button
					onclick={() => {
						this.newName = '';
						onNameChange && onNameChange(newName);
					}}
				>
					Set new name
				</button>
			</div>
		);
	}
}
```

docs/en/creating-widgets/supplemental.md
```diff
@@ -0,0 +1,869 @@
...
+
+```tsx
+import { create, tsx } from '@dojo/framework/core/vdom';
+
+const factory = create().properties<{
+	name: string;
+	onNameChange?(newName: string): void;
+}>();
+
+export default factory(function MyWidget({ properties: { name, onNameChange } }) {
```
agubler: Similar comment to below, the `newName` value needs to persist across renders and be passed into the `input` here, something like this:

```ts
const factory = create({ icache }).properties<{
	name: string;
	onNameChange?(newName: string): void;
}>();

export default factory(function MyWidget({ middleware: { icache }, properties: { name, onNameChange } }) {
	let newName = cache.get<string>('new-name') || '';
	return (
		<div>
			<span>Hello, {name}! Not you? Set your name:</span>
			<input
				type="text"
				value={newName}
				oninput={(e: Event) => {
					icache.set('new-name', (e.target as HTMLInputElement).value);
				}}
			/>
			<button
				onclick={() => {
					icache.set('new-name', undefined);
					onNameChange && onNameChange(newName);
				}}
			>
				Set new name
			</button>
		</div>
	);
});
```

docs/en/creating-widgets/supplemental.md
```diff
@@ -0,0 +1,869 @@
...
+```tsx
+import { create, tsx, invalidator } from '@dojo/framework/core/vdom';
+
+import MyWidget from './MyWidget';
+
+const factory = create({ invalidator });
+
+let currentName: string = 'Alice';
+
+export default factory(function NameHandler({ middleware: { invalidator } }) {
```
agubler: Same comment here, should use `cache` or `icache`

docs/en/creating-widgets/supplemental.md
```diff
@@ -0,0 +1,869 @@
...
+    -   Virtual nodes are designed to be lightweight and cost very little to instantiate new versions of within every widget render cycle.
+    -   The framework relies on having two separate virtual node instances across two widget render function calls to perform accurate change detection. If no changes are detected, no further cost is incurred, rendering or otherwise.
+
+## Rendering and the DOM
+
+-   Applications should have no need to use imperative DOM manipulation calls.
+    -   The framework handles all concrete rendering responsibilities, and provides alternative mechanisms to widget authors for using a variety of DOM functionality in a more simplified, type-safe, and reactive manner.
+-   Dojo does not support changing event listener functions after the first render.
+    -   Only changes to node properties and attributes are reflected against DOM nodes between subsequent render cycles - listener functions that need to vary their implementation between renders should be avoided.
```
agubler: "listener functions that need to vary their implementation between renders should be avoided." I don't think that this is true, a listener can be changed every render.

docs/en/creating-widgets/supplemental.md
```diff
@@ -0,0 +1,869 @@
...
+-   Internal or private state can be fully encapsulated within a widget, if required.
+    -   It is a valid and often desireable pattern to implement 'pure' widgets that incur no side effects and receive their entire state as properties, but this is not the only pattern in Dojo widget development.
+
+## Using class-based widgets
+
+-   The _`__render__`_, _`__setProperties__`_, and _`__setChildren__`_ functions are internal framework implementation details and should never be called nor overridden in application code.
+-   Applications should not instantiate widgets directly - Dojo fully manages the lifecycle of widget instances, including instantiation, caching and destruction.
+    -   Applications only need to provide widgets to the framework as generic type parameters when calling [the `w()` method](#instantiating-vdom-nodes), or implicitly via [TSX tags](#tsx-widget-example), as VDOM output from widgets' render functions.
```
agubler: "Applications only need to provide widgets to the framework as generic type parameters when calling [the `w()` method](#instantiating-vdom-nodes), or implicitly via [TSX tags](#tsx-widget-example), as VDOM output from widgets' render functions."

I'm not sure I understand what this means?

docs/en/creating-widgets/supplemental.md
sbinge: Using `this.invalidate()` was mentioned above these examples:

> For class-based widgets, there are two ways to invalidate:
> 
> 1.  Explicitly calling `this.invalidate()` in an appropriate location where state is being changed.
>     -   In the `MyEncapsulatedStateWidget` example, this could be done in the 'Change State' button's `onclick` handler.

would it be better to swap usage here? `invalidate` in the example and `@watch` as the alternative in the above description?

docs/en/creating-widgets/supplemental.md
sbinge: 👍 more outdated info from the existing docs. Removed.

docs/en/creating-widgets/supplemental.md
sbinge: Convoluted way of saying refer to widgets as generic type arguments, not as concrete instances. I've removed this sub point as the main one is a lot clearer.

sbinge:
@agubler all fixed apart from the query around the `this.invalidate()` vs `@watch()` example.

docs/en/creating-widgets/basic-usage.md
```diff
@@ -0,0 +1,106 @@
...
+```tsx
+import { create, tsx } from '@dojo/framework/core/vdom';
+
+const factory = create().properties<{
+	name: string;
+	onNameChange?(newName: string): void;
+}>();
+
+export default factory(function Greeter({ properties: { name, onNameChange } }) {
+	let newName: string = '';
```
agubler: @sbinge I think I missed this one, does this example also need to use the cache so the values are persisted across renders?

docs/en/creating-widgets/supplemental.md
agubler: I think is fine, given the mentions above

docs/en/creating-widgets/supplemental.md
```diff
@@ -0,0 +1,891 @@
...
+import { tsx } from '@dojo/framework/core/vdom';
+
+export default class MyTsxWidget extends WidgetBase {
+	protected render() {
+		return <div>Hello from a TSX widget!</div>;
+	}
+}
+```
+
+Widgets that need to return multiple top-level TSX nodes can wrap them in a `<virtual>` container element. This is a clearer option than returning an array of nodes as it allows for more natural automated code formatting within TSX blocks. For example:
```
agubler: It's also possible to use `v('virtual',` instead of returning an array - wonder whether we should mention that or not.

docs/en/creating-widgets/basic-usage.md
```diff
@@ -0,0 +1,106 @@
...
+> src/widgets/NameHandler.tsx
+
+```tsx
+import { create, tsx, invalidator } from '@dojo/framework/core/vdom';
+
+import Greeter from './Greeter';
+
+const factory = create({ invalidator });
+
+let currentName: string;
```
agubler: Same comment here, this should not be outside the widget - will need to use `cache` or `icache` to persist the `currentName` across renders and not shared across widget instances

docs/en/creating-widgets/supplemental.md
```diff
@@ -0,0 +1,891 @@
...
+        └── unit
+            └── MyScaffoldedWidget.ts
+
+The location where styles and tests are created can be customized using the `--styles` and `--tests` arguments respectively.
+
+A [Custom Element](https://www.w3.org/TR/2016/WD-custom-elements-20161013/) descriptor will be generated if the `--component` argument is passed. When doing so, the scaffolded widget will also include an appropriate `@customElement` decorator with an empty template object, ready for further configuration.
+
+# Rendering in Dojo
+
+Dojo is a reactive framework, handling responsibilities of data change propagation and associated rendering updates behind the scenes. It uses a virtual DOM (VDOM) concept to represent elements intended for output, with nodes in the VDOM being simple JavaScript objects that are designed to be more efficient to work with than actual DOM elements.
```
dylans:
```suggestion
Dojo is a reactive framework, handling responsibilities of data change propagation and associated rendering updates behind the scenes. Dojo leverages a virtual DOM (VDOM) concept to represent elements intended for output, with nodes in the VDOM being simple JavaScript objects that are designed to be more efficient for developers to work with than actual DOM elements.
```

docs/en/creating-widgets/supplemental.md
```diff
@@ -0,0 +1,891 @@
...
+
+Applications only need to concern themselves with declaring their intended output structure as a hierarchy of virtual DOM nodes, typically done as the return values from their [widgets' render functions](#basic-widget-structure). The framework's [`Renderer`](#rendering-to-the-dom) component then synchronizes the intended output with concrete elements in the DOM. Virtual DOM nodes also serve to configure and provide state to widgets and elements, by passing in properties.
```
dylans:
```suggestion
Applications only need to concern themselves with declaring their intended output structure as a hierarchy of virtual DOM nodes, typically done as the return values from their [widgets' render functions](#basic-widget-structure). The framework's [`Renderer`](#rendering-to-the-dom) component then synchronizes the intended output with concrete elements in the DOM. Virtual DOM nodes also serve to configure and provide state to widgets and elements by passing in properties.
```

docs/en/creating-widgets/supplemental.md
```diff
@@ -0,0 +1,891 @@
...
+    ├── styles/
+    │   ├── myscaffoldedwidget.m.css
+    │   └── myscaffoldedwidget.m.css.d.ts
+    └── tests/
+        └── unit
+            └── MyScaffoldedWidget.ts
+
+The location where styles and tests are created can be customized using the `--styles` and `--tests` arguments respectively.
+
+A [Custom Element](https://www.w3.org/TR/2016/WD-custom-elements-20161013/) descriptor will be generated if the `--component` argument is passed. When doing so, the scaffolded widget will also include an appropriate `@customElement` decorator with an empty template object, ready for further configuration.
```
dylans:
```suggestion
A [Custom Element](https://www.w3.org/TR/2016/WD-custom-elements-20161013/) descriptor will be generated if the `--component` argument is passed.
```

docs/en/creating-widgets/supplemental.md
```diff
@@ -0,0 +1,891 @@
...
+# Rendering in Dojo
+
+Dojo is a reactive framework, handling responsibilities of data change propagation and associated rendering updates behind the scenes. It uses a virtual DOM (VDOM) concept to represent elements intended for output, with nodes in the VDOM being simple JavaScript objects that are designed to be more efficient to work with than actual DOM elements.
+
+Applications only need to concern themselves with declaring their intended output structure as a hierarchy of virtual DOM nodes, typically done as the return values from their [widgets' render functions](#basic-widget-structure). The framework's [`Renderer`](#rendering-to-the-dom) component then synchronizes the intended output with concrete elements in the DOM. Virtual DOM nodes also serve to configure and provide state to widgets and elements, by passing in properties.
+
+Dojo supports subtree rendering, meaning that when a change in state occurs, the framework is able to determine specific subsets of VDOM nodes affected by the change. Only the required corresponding subtrees within are then updated to reflect the change, increasing rendering performance and resulting in an improved user interactive experience.
```
dylans:
```suggestion
Dojo supports subtree rendering, meaning that when a change in state occurs, the framework is able to determine specific subsets of VDOM nodes affected by the change. Only the required corresponding subtrees within the DOM tree are then updated to reflect the change, increasing rendering performance and improving user interactivity and experience.
```

docs/en/creating-widgets/supplemental.md
```diff
@@ -0,0 +1,891 @@
...
+
+> **Note:** Returning virtual nodes from widget render functions is the only concern applications have around rendering. Attempting to use any other practice [is considered an anti-pattern in Dojo application development](#widget-development-best-practices), and should be avoided.
+
+## TSX Support
+
+Dojo supports use of the `jsx` syntax extension known as [`tsx` in TypeScript](https://www.TypeScriptlang.org/docs/handbook/jsx.html). This syntax allows for a more convenient representation of a widget's VDOM output that is closer to the resulting HTML within a built application.
+
+### TSX-enabled applications
+
+TSX-enabled projects can easily be scaffolded via the [`dojo create app --tsx` CLI command](https://github.com/dojo/cli-create-app).
```
dylans:
```suggestion
TSX-enabled projects can easily get scaffolded via the [`dojo create app --tsx` CLI command](https://github.com/dojo/cli-create-app).
```

docs/en/creating-widgets/supplemental.md
```diff
@@ -0,0 +1,891 @@
...
+
+For Dojo projects that were not scaffolded in this way, TSX can be enabled with the following additions to the project's TypeScript config:
+
+> `./tsconfig.json`
+
+```json
+{
+	"compilerOptions": {
+		"jsx": "react",
```
dylans: Note, we should file an enhancement request on the TS team to add a dojo option here. :)

docs/en/creating-widgets/supplemental.md
```diff
@@ -0,0 +1,891 @@
...
+## Working with the VDOM
+
+### VDOM node types
+
+Dojo recognizes two types of nodes within its VDOM:
+
+-   `VNode`s, or _Virtual Nodes_, which are virtual representations of concrete DOM elements, and serve as the lowest-level rendering output for all Dojo applications.
+-   `WNode`s, or _Widget Nodes_, which tie Dojo widgets into the VDOM hierarchy.
+
+Both `VNode`s and `WNode`s are considered subtypes of `DNode`s within Dojo's virtual DOM, but applications don't typically deal with `DNode`s in their abstract sense. Using [TSX syntax](#tsx-support) is also preferred as it allows applications to render both virtual node types with a uniform syntax.
```
dylans:
```suggestion
Both `VNode`s and `WNode`s are considered subtypes of `DNode`s within Dojo's virtual DOM, but applications don't typically deal with `DNode`s in their abstract sense. Using [TSX syntax](#tsx-support) is also preferred as it allows applications to render both virtual node types with uniform syntax.
```

docs/en/creating-widgets/supplemental.md
```diff
@@ -0,0 +1,891 @@
...
+| `children`             | Yes               | An array of nodes to render as children of the newly created node. This can also include any text node children as literal strings, if required. Widgets typically encapsulate their own children, so this argument is more likely to be used with `v()` than `w()`.                                                                                                                                                                                                                                                                           |
+
+### Virtual nodes example
+
+The following sample widget includes a more typical render function that returns a `VNode`. It has an intended structural representation of a simple `div` DOM element, that includes a text child node:
```
dylans:
```suggestion
The following sample widget includes a more typical render function that returns a `VNode`. It has an intended structural representation of a simple `div` DOM element which includes a text child node:
```

docs/en/creating-widgets/supplemental.md
```diff
@@ -0,0 +1,891 @@
...
+
+import MyComposingWidget from './widgets/MyComposingWidget';
+
+const r = renderer(() => <MyComposingWidget />);
+r.mount();
+```
+
+### `MountOptions` properties
+
+The `Renderer.mount()` method accepts an optional `MountOptions` argument that configures how the mount operation is performed.
```
dylans:
```suggestion
The `Renderer.mount()` method accepts an optional `MountOptions` argument that configures how the mount operation gets performed.
```

docs/en/creating-widgets/supplemental.md
```diff
@@ -0,0 +1,891 @@
...
+| `on`       | Yes      | The set of events to apply to the external DOM node                                                                                                                                       |
+| `diffType` | Yes      | Default: `none`. The [change detection strategy](#external-dom-node-change-detection) to use when determining if the external DOM node requires updating from within the Dojo application |
+| `onAttach` | Yes      | An optional callback that is executed after the node has been appended to the DOM                                                                                                         |
+
+### External DOM node change detection
+
+External nodes added through `dom()` are a step removed from regular virtual nodes as its possible for them be managed outside of the Dojo application. This means Dojo can't use the `VNode`'s properties as the master state for the element, but instead has to rely on the underlying javascript properties and HTML attributes on the DOM node itself.
```
dylans:
```suggestion
External nodes added through `dom()` are a step removed from regular virtual DOM nodes as it is possible for them get managed outside of the Dojo application. This means Dojo cannot use the `VNode`'s properties as the master state for the element, but instead must rely on the underlying JavaScript properties and HTML attributes on the DOM node itself.
```

docs/en/creating-widgets/supplemental.md
```diff
@@ -0,0 +1,891 @@
...
+
+`dom()` accepts a `diffType` property that allows users to specify a property change detection strategy for the wrapped node. A particular strategy given the wrapped node's intended usage can help Dojo to determine if a property or attribute has changed, and therefore needs to be applied to the wrapped DOM node. The default strategy is `none`, meaning Dojo will simply add the wrapped DOM element as-is within the application's output on every render cycle.
+
+**Note:** All strategies use the events from the previous `VNode` to ensure that they are correctly removed and applied each render.
```
dylans:
```suggestion
**Note:** All strategies use the events from the previous `VNode` to ensure that they are correctly removed and applied upon each render.
```

docs/en/creating-widgets/supplemental.md
```diff
@@ -0,0 +1,891 @@
...
+| ---------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `none`     | This mode passes an empty object as the previous `attributes` and `properties` within the wrapping `VNode`, meaning the `props` and `attrs` passed to `dom()` will always be reapplied to the wrapped node in every render cycle. |
+| `dom`      | This mode uses the `attributes` and `properties` from the DOM node as the base to calculate if there is a difference from the `props` and `attrs` passed to `dom()` that then need to get applied.                                |
+| `vdom`     | This mode will use the previous `VNode` for the diff, which is effectively Dojo's default VDOM diff strategy. Any changes made directly to the wrapped node will be ignored in terms of change detection and rendering updates.   |
```
dylans:
```suggestion
| `vdom`     | This mode will use the previous `VNode` for the diff, which is effectively Dojo's default VDOM diff strategy. Any changes made directly to the wrapped node will get ignored in terms of change detection and rendering updates.   |
```

docs/en/creating-widgets/supplemental.md
```diff
@@ -0,0 +1,891 @@
...
+
+## Configuring `VNode`s
+
+`VNodeProperties` contains many fields that act as the primary API to interact with concrete elements in the DOM. Many of these properties mirror those available on [`HTMLElement`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement), including specifying various `oneventname` event handlers.
+
+Application of these properties is considered unidirectional, in that Dojo applies the given set to concrete DOM elements but does not synchronize any further changes made to the corresponding DOM attributes back into the given `VNodeProperties`. Any such changes should instead be propagated back into the Dojo application via event handlers. When an event handler is invoked, the application can process any change in state required for the event, update its view of the corresponding `VNodeProperties` when outputting its VDOM structure for rendering, then let Dojo's [`Renderer` synchronize any relevant updates with the DOM](#rendering-to-the-dom).
```
dylans:
```suggestion
Application of these properties is considered unidirectional, in that Dojo applies the given set to concrete DOM elements but does not synchronize any further changes made to the corresponding DOM attributes back into the given `VNodeProperties`. Any such changes should instead be propagated back into the Dojo application via event handlers. When an event handler is invoked, the application can process any change in state required for the event, update its view of the corresponding `VNodeProperties` when outputting its VDOM structure for rendering, and then let Dojo's [`Renderer` synchronize any relevant updates with the DOM](#rendering-to-the-dom).
```

docs/en/creating-widgets/supplemental.md
```diff
@@ -0,0 +1,891 @@
...
+
+## Changing properties and diff detection
+
+Dojo uses virtual node properties to determine if a given node has been updated and therefore requires re-rendering. Specifically, it uses a difference detection strategy to compare sets of properties from the previous and current render frames. If a difference is detected in the latest set of properties that a node receives, that node is invalidated and will be re-rendered in the next paint cycle.
```
dylans:
```suggestion
Dojo uses virtual node properties to determine if a given node has been updated and therefore requires re-rendering. Specifically, it uses a difference detection strategy to compare sets of properties from the previous and current render frames. If a difference is detected in the latest set of properties that a node receives, that node is invalidated and gets re-rendered in the next paint cycle.
```

docs/en/creating-widgets/supplemental.md
```diff
@@ -0,0 +1,891 @@
...
+}
+```
+
+### Delegating focus
+
+Function-based widgets can use the [`focus` middleware](../middleware/supplemental.md#focus) to provide `focus` to their children or to accept `focus` from a parent widget.
+
+Class-based widgets can use the `FocusMixin` (from `@dojo/framework/core/mixins/Focus`) to delegate focus in a similar way.
+
+The `FocusMixin` adds `focus` and `shouldFocus` to a widget's API. `shouldFocus` checks if the widget is in a state to perform a focus action and will only return `true` once, until the widget's `focus` method has been called again. This `shouldFocus` method is designed to be passed to child widgets or nodes as the value of their `focus` property.
```
dylans:
```suggestion
The `focus` middleware or `FocusMixin` adds `focus` and `shouldFocus` to a widget's API. `shouldFocus` checks if the widget is in a state to perform a focus action and will only return `true` once, until the widget's `focus` method has been called again. This `shouldFocus` method is designed to be passed to child widgets or nodes as the value of their `focus` property.
```

docs/en/creating-widgets/supplemental.md
```diff
@@ -0,0 +1,891 @@
...
+
+When `shouldFocus` is passed to a widget, it will be called as the properties are set on the child widget, meaning that any other usages of the parent's `shouldFocus` method will result in a return value of `false`.
+
+An example usage controlling focus across child VNodes (DOM) and WNodes (widgets):
+
+> src/widgets/FocusableWidget.tsx
+
+```tsx
```
dylans: Do we need a function-based variant example here?

docs/en/creating-widgets/supplemental.md
```diff
@@ -0,0 +1,891 @@
...
+				/>
+			</div>
+		);
+	}
+}
+```
+
+# Managing state
+
+For simple applications where data is not required to flow between many components, state management can be straightforward to deal with. Data can be encapsulated within individual widgets that need it as the most [basic form of state management](#basic-internal-widget-state) within a Dojo application.
```
dylans:
```suggestion
For simple applications where data is not required to flow between many components, state management can be very straightforward. Data can be encapsulated within individual widgets that need it as the most [basic form of state management](#basic-internal-widget-state) within a Dojo application.
```

docs/en/creating-widgets/supplemental.md
```diff
@@ -0,0 +1,891 @@
...
+
+For class-based widgets, there are two ways to invalidate:
+
+1.  Explicitly calling `this.invalidate()` in an appropriate location where state is being changed.
+    -   In the `MyEncapsulatedStateWidget` example, this could be done in the 'Change State' button's `onclick` handler.
+2.  Annotating any relevant fields with the `@watch()` decorator (from the `@dojo/framework/core/vdomecorators/watch` module). When `@watch`ed fields are modified, `this.invalidate()` will implicitly be called - this can be useful for state fields that always need to trigger a re-render when updated.
+
+**Note:** marking a widget as invalid won't immediately re-render the widget - instead it acts as a notification to Dojo that the widget has been dirtied, so should be updated and re-rendered in the next render cycle. This means invalidating a widget multiple times within the same render frame won't have any negative impact on application performance, although excessive invalidation should be avoided to ensure optimal performance.
```
dylans:
```suggestion
**Note:** marking a widget as invalid won't immediately re-render the widget - instead it acts as a notification to Dojo that the widget is in a dirty state and should be updated and re-rendered in the next render cycle. This means invalidating a widget multiple times within the same render frame won't have a negative impact on application performance, although excessive invalidation should be avoided to ensure optimal performance.
```

docs/en/creating-widgets/supplemental.md
```diff
@@ -0,0 +1,891 @@
...
+				}}
+			/>
+		);
+	}
+}
+```
+
+## Advanced: abstracting and injecting state
+
+When implementing complex responsibilities, following a pattern of state encapsulation within widgets can result in bloated, unmanageable components. Another problem can arise in large applications where state is required in leaf widgets, but not in intermediate containers within the VDOM hierarchy. Passing state down such a complex widget hierarchy adds brittle, unnecessary code.
```
dylans: It would be nice if we could suggest what large means.

docs/en/creating-widgets/basic-usage.md
```diff
@@ -0,0 +1,109 @@
...
+				icache.set('current-name', newName);
+			}}
+		/>
+	);
+});
+```
+
+## Rendering to the DOM
+
+-   Connecting a widget hierarchy to its concrete representation within a webpage
```
agubler: Trying to think of a clearer way to word this, sounds a bit complicated but it isn't.

docs/en/creating-widgets/supplemental.md
```diff
@@ -0,0 +1,972 @@
...
+
+Because this widget returns an empty array from its render function, it has no structural representation within an application's output. Widgets typically [return one or more virtual DOM nodes](#virtual-nodes-example) in order to provide meaningful structure within the application's HTML output.
+
+The process of translating virtual DOM nodes to output on a web page is handled by [Dojo's rendering system](#rendering-in-dojo).
+
+## Widget styling
+
+Styling of a widget's DOM output is handled via CSS, with relevant style classes stored in a CSS module file parallel to the widget's TypeScript module. Styling is identical for both function- and class-based widget variants. This topic is described in detail within the [Styling and Theming reference guide](../styling-and-theming/supplemental.md).
+
+## Scaffolding widgets
```
agubler: I'm beginning to think we don't need this section in the guide... what do others think?

docs/en/creating-widgets/supplemental.md
```diff
@@ -0,0 +1,972 @@
...
+    ├── styles/
+    │   ├── myscaffoldedwidget.m.css
+    │   └── myscaffoldedwidget.m.css.d.ts
+    └── tests/
+        └── unit
+            └── MyScaffoldedWidget.ts
+
+The location where styles and tests are created can be customized using the `--styles` and `--tests` arguments respectively.
+
+A [Custom Element](https://www.w3.org/TR/2016/WD-custom-elements-20161013/) descriptor will be generated if the `--component` argument is passed.
```
agubler: If we do keep this section, this should be removed as this has been streamlined so that the custom element decorator is not needed to build a custom element.

docs/en/creating-widgets/supplemental.md
```diff
@@ -0,0 +1,972 @@
...
+
+> `./tsconfig.json`
+
+```json
+{
+	"compilerOptions": {
+		"jsx": "react",
+		"jsxFactory": "tsx"
+	},
+	"include": ["./src/**/*.ts", "./src/**/*.tsx"]
```
agubler: probably should include the `test` directories too.

docs/en/creating-widgets/supplemental.md
```diff
@@ -0,0 +1,972 @@
...
+**Function-based variant:**
+
+```ts
+import { create, v, w } from '@dojo/framework/core/vdom';
+
+const factory = create();
+
+import MyWidget from './MyWidget';
+
+export default factory(function MyComposingWidget() =>
```
agubler: This is not valid syntax with the `=>` - it will need to be 
```ts
export default factory(function MyComposingWidget() {
    return v('div', ['This widget outputs several virtual nodes in a hierarchy', w(MyWidget, {})]);
});
```

docs/en/creating-widgets/supplemental.md
```diff
@@ -0,0 +1,972 @@
...
+| ---------- | ------------- | -------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `sync`     | `boolean`     | Yes      | Default: `false`. If `true`, relevant render lifecycle callbacks (specifically, `after` and `deferred` render callbacks) are run synchronously. If `false`, the callbacks are instead scheduled to run asynchronously before the next repaint via [`window.requestAnimationFrame()`](https://developer.mozilla.org/en-US/docs/Web/API/window/requestAnimationFrame). Synchronous render callbacks can be beneficial in scenarios where specific nodes need to exist in the DOM. |
```
agubler: I think we should mention, sync is not recommended for standard applications.

docs/en/creating-widgets/supplemental.md
```diff
@@ -0,0 +1,972 @@
...
+```
+
+> src/main.tsx
+
+```tsx
+import renderer, { tsx } from '@dojo/framework/core/vdom';
+
+import MyComposingWidget from './widgets/MyComposingWidget';
+
+const dojoAppRootElement = document.getElementById('my-dojo-app') || undefined;
```
agubler: note to me: we should consider adding `null` to the typing of `domNode`

docs/en/creating-widgets/supplemental.md
agubler: I think that would be in the middleware guide.

docs/en/creating-widgets/supplemental.md
sbinge: Missed this one when converting over to named functions.

docs/en/creating-widgets/supplemental.md
sbinge: Removed in favor of an upcoming Getting Started guide

docs/en/creating-widgets/basic-usage.md
sbinge: Tried a new sentence. hopefully slightly clearer now

sbinge:
@agubler this should hopefully be good to go now

PR #385, opened 2019-06-07
Intersection Middleware

### LINKED ISSUE #374: Intersection Middleware
**Enhancement**

An intersection middleware that enables user to determine if a node is visible within the view port using [Intersection Observers](https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API).

The middleware should at least have feature parity with the existing intersection meta.

agubler:
**Type:** feature

**Description:**

Middleware that leverages Intersection Observers to determine if a node with a given key is intersecting with the view port or with a specified root node.

Resolves #374

### DIFF
--- a/tests/core/support/jsdom-plugin.ts
+++ b/tests/core/support/jsdom-plugin.ts
@@ -67,6 +67,7 @@ intern.registerPlugin('jsdom', async () => {
 			}
 		});
 	};
+	initialize();
 
 	const uninitialize = () => {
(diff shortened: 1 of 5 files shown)


tests/core/support/jsdom-plugin.ts
```diff
@@ -67,6 +67,7 @@ intern.registerPlugin('jsdom', async () => {
 			}
 		});
 	};
+	initialize();
```
agubler: This is because jsdom needs to be initialised for when the IO polyfill is required in order to run the tests in node. I think we should simply always initialise jsdom for testing rather than requiring a customised "describe" that implicitly performs an initialisation for tests it contains, but that should probably be a separate change.

PR #391, opened 2019-06-07
Store Middleware

### LINKED ISSUE #397: Store Middleware
**Enhancement**

A middleware to interact with a store. Providing APIs for `get`, `at`, `path` that reflect the APIs from the `Store`. On use of `get` the middleware should automatically subscribe the widget to data changes (by default, with an option to turn that off).

Also requires an API that wires a process to the store.

Lastly, by default the middleware should factory a Store that is shared across all the middleware instances, but use a store from the registry if one exists for backwards compatibility.

agubler:
**Type:** feature

**Description:**

Dojo stores middleware

Resolves #397

### DIFF
--- a/src/core/middleware/store.ts
+++ b/src/core/middleware/store.ts
@@ -1,0 +1,45 @@
+import { destroy, invalidator, create } from '../vdom';
+import Store, { StatePaths, Path } from '../../stores/Store';
+import { Process } from '../../stores/process';
+
+const factory = create({ destroy, invalidator });
+
+export const createStoreMiddleware = <S = any>(initial?: Function) => {
+	let store = new Store();
+	initial && initial(store);
+	const storeMiddleware = factory(({ middleware: { destroy, invalidator } }) => {
+		const handles: any[] = [];
+		destroy(() => {
+			let handle: any;
+			while ((handle = handles.pop())) {
+				handle();
+			}
+		});
+		const registeredPaths: string[] = [];
+		const path: StatePaths<S> = (path: any, ...segments: any) => {
+			return (store as any).path(path, ...segments);
+		};
+		return {
+			get<U = any>(path: Path<S, U>): U {
+				if (registeredPaths.indexOf(path.path) === -1) {
+					const handle = store.onChange(path, () => {
+						invalidator();
+					});
+					handles.push(() => handle.remove());
+					registeredPaths.push(path.path);
+				}
+				return store.get(path);
+			},
+			path,
+			at<U = any>(path: Path<S, U[]>, index: number) {
+				return store.at(path, index);
+			},
+			executor<T extends Process<any, any>>(process: T): ReturnType<T> {
+				return process(store) as any;
+			}
+		};
+	});
+	return storeMiddleware;
+};
+
+export default createStoreMiddleware();
--- a/src/core/middleware/store.ts
+++ b/src/core/middleware/store.ts
@@ -1,13 +1,15 @@
 import { destroy, invalidator, create } from '../vdom';
+import injector from '../middleware/injector';
 import Store, { StatePaths, Path } from '../../stores/Store';
 import { Process } from '../../stores/process';
 
-const factory = create({ destroy, invalidator });
+const factory = create({ destroy, invalidator, injector });
 
-export const createStoreMiddleware = <S = any>(initial?: Function) => {
-	let store = new Store();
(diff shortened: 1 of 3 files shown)


src/core/middleware/store.ts
```diff
@@ -0,0 +1,45 @@
+import { destroy, invalidator, create } from '../vdom';
+import Store, { StatePaths, Path } from '../../stores/Store';
+import { Process } from '../../stores/process';
+
+const factory = create({ destroy, invalidator });
+
+export const createStoreMiddleware = <S = any>(initial?: Function) => {
```
agubler: Supports running an optional initial set of processes to bootstrap the store data

agubler: The factory is used to type the store middleware for the application state

src/core/middleware/store.ts
```diff
@@ -0,0 +1,45 @@
...
+	let store = new Store();
```
agubler: The store is factory'd for the middleware, does not use the store from the injector - We might want to support the injector from the store for backwards compatibility (so that users can progressively adopt functional widgets and middleware)

src/core/middleware/store.ts
```diff
@@ -0,0 +1,45 @@
...
+			while ((handle = handles.pop())) {
+				handle();
+			}
+		});
+		const registeredPaths: string[] = [];
+		const path: StatePaths<S> = (path: any, ...segments: any) => {
+			return (store as any).path(path, ...segments);
+		};
+		return {
+			get<U = any>(path: Path<S, U>): U {
```
agubler: Automatically registers paths to invalidate on for a widget

src/core/middleware/store.ts
```diff
@@ -0,0 +1,45 @@
...
+					handles.push(() => handle.remove());
+					registeredPaths.push(path.path);
+				}
+				return store.get(path);
+			},
+			path,
+			at<U = any>(path: Path<S, U[]>, index: number) {
+				return store.at(path, index);
+			},
+			executor<T extends Process<any, any>>(process: T): ReturnType<T> {
```
agubler: function that returns an executor from a process, `const exec = executor(process)` is equivalent to `const exec = process(store);`

src/core/middleware/store.ts
```diff
@@ -0,0 +1,54 @@
...
+				return process(store) as any;
+			}
+		};
+	});
+	return storeMiddleware;
+};
+
+export default createStoreMiddleware;
```
agubler: Exports a factory that creates the middleware, this is so that it can be typed with the store state interface, also factories a Store that is shared across all instances of the middleware.

src/core/middleware/store.ts
```diff
@@ -0,0 +1,54 @@
+import { destroy, invalidator, create } from '../vdom';
+import injector from '../middleware/injector';
+import Store, { StatePaths, Path } from '../../stores/Store';
+import { Process } from '../../stores/process';
+
+const factory = create({ destroy, invalidator, injector });
+
+export const createStoreMiddleware = <S = any>(initial?: (store: Store<S>) => void) => {
```
agubler: Accepts an optional `initial` (perhaps should be called `init` or `initialize`) function that receives the factorie'd store and can initialize state etc.

src/core/middleware/store.ts
```diff
@@ -0,0 +1,54 @@
...
+	initial && initial(store);
+	const storeMiddleware = factory(({ middleware: { destroy, invalidator, injector } }) => {
+		const handles: any[] = [];
+		destroy(() => {
+			let handle: any;
+			while ((handle = handles.pop())) {
+				handle();
+			}
+		});
+		if (!initialized) {
```
agubler: Does a one time check for a store in the application registry, if it exists the middleware will use that instead

src/core/middleware/store.ts
```diff
@@ -0,0 +1,54 @@
...
+				store = injectedStore;
+			}
+			initialized = true;
+		}
+		const registeredPaths: string[] = [];
+		const path: StatePaths<S> = (path: any, ...segments: any) => {
+			return (store as any).path(path, ...segments);
+		};
+		return {
+			get<U = any>(path: Path<S, U>, subscribe = true): U {
```
agubler: Equivalent to the store's `get` API expect it will automatically subscribe the widget to data updates for all the paths used. This can be manually disabled on a call by call basis by passing `false` as the second argument

PR #392, opened 2019-06-07
Theme Middleware

### LINKED ISSUE #371: Theme Middleware
**Enhancement**

A middleware for working with Dojo themes, which includes returning the correct class names based on the registered themes, registering available themes and change an application theme.

agubler:
**Type:** feature

**Description:**

Theme middleware to work to determine the correct classes based on the current theme & set themes.

Resolves #371

### DIFF
--- a/src/core/RegistryHandler.ts
+++ b/src/core/RegistryHandler.ts
@@ -11,7 +11,7 @@ export class RegistryHandler extends Evented<RegistryHandlerEventMap> {
 	private _registry = new Registry();
 	private _registryWidgetLabelMap: Map<Registry, RegistryLabel[]> = new Map();
 	private _registryInjectorLabelMap: Map<Registry, RegistryLabel[]> = new Map();
-	protected baseRegistry?: Registry;
+	protected baseRegistry!: Registry;
 
 	constructor() {
 		super();
@@ -20,7 +20,7 @@ export class RegistryHandler extends Evented<RegistryHandlerEventMap> {
 			if (this.baseRegistry) {
 				this._registryWidgetLabelMap.delete(this.baseRegistry);
 				this._registryInjectorLabelMap.delete(this.baseRegistry);
-				this.baseRegistry = undefined;
+				this.baseRegistry = undefined as any;
 			}
 		};
 		this.own({ destroy });
@@ -32,6 +32,10 @@ export class RegistryHandler extends Evented<RegistryHandlerEventMap> {
 			this._registryInjectorLabelMap.delete(this.baseRegistry);
 		}
 		this.baseRegistry = baseRegistry;
+	}
+
+	public get base(): Registry {
+		return this.baseRegistry;
 	}
 
 	public define(label: RegistryLabel, widget: RegistryItem): void {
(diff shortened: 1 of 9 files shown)


src/core/middleware/theme.ts
```diff
@@ -0,0 +1,83 @@
+import { create, invalidator, diffProperties, injector } from '../vdom';
+import cache from './cache';
+import { SupportedClassName } from '../interfaces';
+import Injector from '../Injector';
+
+export interface ClassNames {
```
agubler: These interfaces should be reused from the existing `Themed` class

src/core/middleware/theme.ts
```diff
@@ -0,0 +1,83 @@
...
+					const classKey = classKeys[i];
+					if (theme[classKey]) {
+						theme[classKey] = `${theme[classKey]} ${properties.classes[key][classKey].join(' ')}`;
+					}
+				}
+			}
+			cache.set(css, theme);
+			return theme;
+		},
+		set(css?: Theme) {
```
agubler: This API is confusing as it's not the opposite of `get`, it actually sets the current theme - also need an API to register a theme, registering a theme would make it available and only set it if no theme is currently set.

src/core/RegistryHandler.ts
```diff
@@ -34,6 +34,10 @@ export class RegistryHandler extends Evented<RegistryHandlerEventMap> {
 		this.baseRegistry = baseRegistry;
 	}
 
+	public get base(): Registry {
```
agubler: Exposes the base registry - this is the app level repository.

src/core/vdom.ts
```diff
@@ -813,7 +813,7 @@ export function renderer(renderer: () => RenderResult): Renderer {
 		merge: true,
 		transition: transitionStrategy,
 		domNode: global.document.body,
-		registry: null
+		registry: new Registry()
```
agubler: Add a `Registry` by default to the application if one is not passed by the user - this means things like the `theme` middleware will work without passing a registry to the `mount`

src/core/diff.ts
```diff
@@ -26,7 +26,7 @@ export function reference(previousProperty: any, newProperty: any): PropertyChan
 	};
 }
 
-export function shallow(previousProperty: any, newProperty: any): PropertyChangeRecord {
+export function shallow(previousProperty: any, newProperty: any, depth = 0): PropertyChangeRecord {
```
agubler: Adds support to choose the depth of the diff... defaults to 0 which is the equivalent of the current behaviour

src/core/middleware/theme.ts
```diff
@@ -0,0 +1,99 @@
...
+				current.theme !== next.theme ||
+				(current.classes && !next.classes) ||
+				(!current.classes && next.classes)
+			) {
+				result = true;
+			} else if (current.classes && next.classes) {
+				const keys = [...themeKeys.values()];
+				for (let i = 0; i < keys.length; i++) {
+					let key = keys[i];
+					result = shallow(current.classes[key], next.classes[key], 1).changed;
```
agubler: Diffs the classes for every key that has been registered by the widget, the classes object is nested three levels so requires a slightly deeper diff than the standard `shallow` diff. To support this `shallow` accepts a depth param to control the level of the diff.

PR #394, opened 2019-06-10
Fix custom element initialization

### LINKED ISSUE #342: Buggy custom element behavior when script is included before element is used
**Bug** 

Including the `<script>` tag for the dojo button element before the element is referenced, as in this snippet:

```html
<!DOCTYPE html>
<html>
<head>
	<link rel="stylesheet" href="output/dist/button-6.0.0-pre.css" />
	<script src="output/dist/button-6.0.0-pre.js"></script>
</head>
<body>
<dojo-button>Hello</dojo-button>
</body>
</html>
```

Will cause the text Hello to not be inserted correctly as a child element.

image

Moving the script tag to the bottom, as in:

```html
<!DOCTYPE html>
<html>
<head>
	<link rel="stylesheet" href="output/dist/button-6.0.0-pre.css" />
</head>
<body>
<dojo-button>Hello</dojo-button>
<script src="output/dist/button-6.0.0-pre.js"></script>
</body>
</html>
```

Results in the proper DOM structure.

image

Package Version: 6.0.0-alpha.1

**Code**

I just built the @dojo/widgets custom elements and put the previously posted html file(s) in the root.

**Expected behavior:**

Custom elements will be initialized correctly regardless of whether or not their script has been loaded first.

**Actual behavior:**

Custom elements are only initialized correctly if their script is loaded after the element has been used.

maier49:
**Type:** bug

**Description:**
It's evidently not guaranteed that a custom element's children will be available when it is connected ([see discussion here](https://github.com/w3c/webcomponents/issues/551)), but our `connectedCallback` was setup as if it was. I've setup a mutation observer so that the element will wait till the document is loaded  or an element after it in the HTML hierarchy is present to perform setup logic. 

Resolves #342

### DIFF
--- a/tests/core/unit/registerCustomElement.ts
+++ b/tests/core/unit/registerCustomElement.ts
@@ -40,6 +40,28 @@ class DisplayElementDefault extends WidgetBase {
 	}
 }
 
+@customElement({
+	tag: 'delayed-children-element',
+	childType: CustomElementChildType.TEXT
+})
+class DelayedChildrenWidget extends WidgetBase {
+	render() {
+		return v(
+			'div',
+			{},
+			this.children.map((child, i) =>
+				v(
+					'div',
+					{
+						'data-key': `child-${i}`
+					},
+					[child]
+				)
+			)
+		);
+	}
+}
+
 function createTestWidget(options: any) {
 	const { properties, attributes, events, childType = CustomElementChildType.DOJO } = options;
 	@customElement<any>({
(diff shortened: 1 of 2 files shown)


agubler:
@maier49 FYI, this seems to be failing CI.

tests/core/unit/registerCustomElement.ts
```diff
@@ -349,4 +372,35 @@ describe('registerCustomElement', () => {
...
+		Object.defineProperty(document, 'readyState', {
+			configurable: true,
+			get() {
+				return 'complete';
+			}
+		});
+		child = document.createTextNode('bar');
+		element!.appendChild(child);
+
+		waitFor(
```
agubler: curious, do you need to return this to ensure that the test waits for the promise to be resolved?

maier49: Yes 🤦‍♂ . This is actually broken in IE. Looking into it

maier49: @agubler Ok it was still the test that was broken I just didn't fix it properly. Should be good now (pending CI)

maier49:
@agubler A problem with the approach I was using is that if no children are appended it might never call the ready callback. I've updated it to use `setTimeout`.

maier49:
@agubler Testing against the v5 custom elements showcase everything worked properly out of the box. But there were was an issue with the setup being called too many times when I put the element scripts in the head. Have that fixed now, here's the showcase with the scripts in the head:
https://dist-ytwugedbhm.now.sh/

PR #405, opened 2019-06-21
Fix store `state` interface call stack error

### LINKED ISSUE #404: Maximum call stack exceeded error working with stores `state`
**Bug**

An error is occasionally thrown when working with arrays using the new `state` API in Dojo Stores commands.

```
Pointer.ts:76 Uncaught RangeError: Maximum call stack size exceeded
    at Array.map (<anonymous>)
    at new Pointer (Pointer.ts:76)
    at DefaultState.path (Store.ts:160)
    at Object.get (process.ts:219)
    at Pointer.get (Pointer.ts:92)
    at DefaultState.path (Store.ts:170)
    at Object.get (process.ts:219)
    at Pointer.get (Pointer.ts:92)
    at DefaultState.path (Store.ts:170)
    at Object.get (process.ts:219)
```

An example of the error can be seen [here](https://dev-jrxlvmhenq.now.sh), steps to reproduce:

1) Add 3 todos
2) Mark the second todo as complete
3) Click the "clear completed" button
4) Open the console and see the call stack error

The source repository for the reproduction can be found [here](https://github.com/agubler/todomvc).

The command that cause is the issue:

```ts
const clearCompletedCommand = commandFactory(({ state }) => {
	if (state.todos) {
		state.todos = state.todos.filter((todo) => !todo.completed);
	}
	state.completedCount = 0;
});

```

maier49:
**Type:** bug

**Description:**
Any object and array values returned when accessing the state object are also proxies, in order to handle nested updates. When one of these values is set back on the state object there is a check to get the object value instead of the proxy. However in the case of `filter`, and any number of other possible scenarios, the object that is being set is not a proxy and is not known by the proxy, but contains proxies. This resulted in proxies being set as properties on the state which leads to an infinite regress.

To fix this the proxy removal check is now recursive. 

todo-mvc with fix:
https://dev-kxynqtvhqn.now.sh/

Fork:
https://github.com/maier49/todomvc/tree/fixed
Resolves #404

### DIFF
--- a/src/stores/process.ts
+++ b/src/stores/process.ts
@@ -168,7 +168,15 @@ export function createCommandFactory<T, P extends object = DefaultPayload>(): Co
 export type Commands<T = any, P extends object = DefaultPayload> = (Command<T, P>[] | Command<T, P>)[];
 
 const processMap = new Map();
-export const valueSymbol = Symbol('value');
+const valueSymbol = Symbol('value');
+
+export function isStateProxy(value: any) {
+	if (typeof value !== 'object' || value === null) {
+		return false;
+	}
+
+	return Boolean(value[valueSymbol]);
+}
 
 export function getProcess(id: string) {
 	return processMap.get(id);
--- a/src/stores/process.ts
+++ b/src/stores/process.ts
@@ -182,17 +182,13 @@ function removeProxies(value: any) {
 			value = value[valueSymbol];
 		}
 
-		if (Array.isArray(value)) {
-			value = value.map(removeProxies);
-		} else {
-			value = Object.keys(value).reduce(
-				(newValue, key) => {
-					newValue[key] = removeProxies(value[key]);
-					return newValue;
-				},
-				{} as typeof value
-			);
+		const newValue: typeof value = Array.isArray(value) ? [] : {};
+		const keys = Object.keys(value);
+		for (let i = 0; i < keys.length; i++) {
+			newValue[keys[i]] = removeProxies(value[keys[i]]);
 		}
+
+		value = newValue;
 	}
 
 	return value;
--- a/src/stores/process.ts
+++ b/src/stores/process.ts
@@ -168,13 +168,35 @@ export function createCommandFactory<T, P extends object = DefaultPayload>(): Co
 export type Commands<T = any, P extends object = DefaultPayload> = (Command<T, P>[] | Command<T, P>)[];
 
 const processMap = new Map();
-const valueSymbol = Symbol('value');
+export const valueSymbol = Symbol('value');
 
 export function getProcess(id: string) {
 	return processMap.get(id);
 }
 
 const proxyError = 'State updates are not available on legacy browsers';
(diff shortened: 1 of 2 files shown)


src/stores/process.ts
```diff
@@ -168,14 +168,36 @@ export function createCommandFactory<T, P extends object = DefaultPayload>(): Co
...
-const valueSymbol = Symbol('value');
+export const valueSymbol = Symbol('value');
 
 export function getProcess(id: string) {
 	return processMap.get(id);
 }
 
 const proxyError = 'State updates are not available on legacy browsers';
 
+function removeProxies(value: any) {
```
agubler: This could be premature, but I tend to avoid using `map` and `reduces` on what could be hot paths with a significant number of keys/items. Would it be possible to use a standard loop? Does the it need to be immutable? Or would it be fine to mutate the existing object? I was thinking something like this, which I think could be made immutable if required.

```ts
	if (typeof value === 'object' && value !== null) {
		if (value[valueSymbol]) {
			value = value[valueSymbol];
		}
		const keys = Object.keys(value);
		for (let i = 0; i < keys.length; i++) {
			value[keys[i]] = removeProxies(value[keys[i]]);
		}
	}
	return value;
```

maier49: This object could be a user created object that was passed in, so we would not want to mutate it. I can use a loop to build the object though.

maier49: 👍 Updated to use a standard loop without mutating the arg

src/stores/process.ts
```diff
@@ -168,14 +168,32 @@ export function createCommandFactory<T, P extends object = DefaultPayload>(): Co
 export type Commands<T = any, P extends object = DefaultPayload> = (Command<T, P>[] | Command<T, P>)[];
 
 const processMap = new Map();
-const valueSymbol = Symbol('value');
+export const valueSymbol = Symbol('value');
```
matt-gadd: is there any way we can test this without exporting the symbol? seems like we're exporting an implementation detail?

maier49: I'm not sure what we could do without leaking something out here. I agree it's not great, but Proxies are not supposed to expose that they are proxies, and it will look upon inspection just like the object it's proxying in this case.

maier49: Would it be preferable if I exposed an `isProxy` check or something that kept the symbol internal? Or created a proxy util module that created the proxy and provided that `isProxy` check that both still used? I know that's still exposing some kind of implementation detail but at least they wouldn't have access to the symbol.

matt-gadd: yeah that might be nicer (something like `isStateProxy()`?, at least people wouldn't be able to tag things with the symbol then. maybe i'm just being overly cautious in the first place.

maier49: Doesn't seem overly cautious to me. Exposing that symbol does open up the user to doing some bad things if they wanted to.

I've made that change now.

PR #410, opened 2019-06-24
Allow returning a node from decorate modifier to replace a node

### LINKED ISSUE #298: Allow decorate to return a new node
**Enhancement**
It would be nice if the `decorate()` method in https://github.com/dojo/framework/blob/master/src/widget-core/d.ts allowed returning a new/different node, this would allow easily switching nodes in a tree.

maier49:
**Type:** feature

**Description:**
Modifies `decorate` so that returning a node or value from the modifier function will replace it in the tree. It excludes `undefined` from the type so we can identify if something was actually returned.
Resolves #298

### DIFF
--- a/src/core/util.ts
+++ b/src/core/util.ts
@@ -373,12 +373,17 @@ export function decorate(
 		shouldDrain = true;
 	}
 
+	function modifyNode(modifier: Modifier<DNode>, node: DNode) {
+		const modifiedNode = modifier(node, breaker);
+		return modifiedNode === undefined ? node : modifiedNode;
+	}
+
 	let nodes = Array.isArray(dNodes) ? [...dNodes] : [dNodes];
 	for (let i = 0; i < nodes.length; i++) {
 		const node = nodes[i];
 		if (node && node !== true) {
 			if (!predicate || predicate(node)) {
-				nodes[i] = modifier(node, breaker) || node;
+				nodes[i] = modifyNode(modifier, node);
 			}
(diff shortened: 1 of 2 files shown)


src/core/util.ts
```diff
@@ -1,12 +1,12 @@
...
-import { DNode, RenderResult } from './interfaces';
+import { DNode, RenderResult, VNode, WNode } from './interfaces';
 import { isWNode, isVNode } from './vdom';
 
 const slice = Array.prototype.slice;
 const hasOwnProperty = Object.prototype.hasOwnProperty;
 
 export interface Modifier<T extends DNode> {
-	(dNode: T, breaker: () => void): void;
+	(dNode: T, breaker: () => void): void | VNode | WNode | null | string | boolean;
```
matt-gadd: should this not just be `void | DNode`?

maier49: The reason I didn't do that was because it's checking if it's a new node based on whether the return is strictly equal to `undefined`. `DNode` includes undefined so if they returned `undefined` as a new node it wouldn't be picked up.

If that's a problem I could add another callback to pass the new node to instead or something to avoid the ambiguity?

matt-gadd: Hmmm, it seems like we'd want to support being able to return `undefined` right?

maier49:
After some more discussion, this functionality is going to be implemented as a separate function rather than being added to the current decorate functionality.

PR #419, opened 2019-06-26
Make middleware argument mandatory for create factory

agubler:
**Type:** bug

**Description:**

In TypeScript 3.4.5, defaulting the middleware argument to `{}` works and `properties` are correctly inferred from middleware and the chained `.properties<{ props: number }>()` function.

However in 3.5.1, this inference breaks when the middleware argument is defaulted, this change makes the middleware argument mandatory (until we can find a better way...).

```ts
const factory = create();
```
becomes 
```ts
const factory = create({});
```


agubler:
closing in favour of #420

PR #423, opened 2019-06-27
Do not queue an invalidation when an invalidate occurs during a render

agubler:
**Type:** bug

**Description:**

From realworld testing some small issue have been identified and resolved:

* Do not add a widget on the invalidation queue when the invalidate occurs whilst "rendering"
* Do not run deferred properties if the owning widget has been destroyed
* Revert change for `Outlet` that moved `invalidate` inside the property diff condition, since it always needs to re-render from the parent.

### DIFF
--- a/src/core/vdom.ts
+++ b/src/core/vdom.ts
@@ -74,6 +74,7 @@ export interface WidgetMeta {
 	registryHandler: RegistryHandler;
 	properties: any;
 	children?: DNode[];
+	rendering: boolean;
 	nodeMap: Map<string | number, HTMLElement>;
 	destroyMap: Map<string, () => void>;
 	deferRefs: number;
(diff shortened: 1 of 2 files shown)


src/routing/Outlet.ts
```diff
@@ -37,8 +37,8 @@ export const Outlet = factory(function Outlet({
 			if (handle) {
 				cache.set('handle', handle);
 			}
-			invalidator();
 		}
+		invalidator();
```
agubler: This needs to `invalidate` every render, but not actually queue an invalidation just mark the widget as dirty.

src/core/vdom.ts
```diff
@@ -1682,13 +1685,15 @@ export function renderer(renderer: () => RenderResult): Renderer {
 					const widgetMeta = widgetMetaMap.get(next.id);
 					if (widgetMeta) {
 						widgetMeta.dirty = true;
+						if (!widgetMeta.rendering && _idToWrapperMap.has(next.id)) {
```
agubler: Only queue an invalidation when not rendering and the widget is still rendered in the page

src/core/vdom.ts
```diff
@@ -1022,11 +1023,13 @@ export function renderer(renderer: () => RenderResult): Renderer {
 		if (deferredPropertiesCallback) {
 			const properties = next.node.properties;
 			_deferredRenderCallbacks.push(() => {
-				const deferredProperties = next.deferredProperties;
-				next.deferredProperties = deferredPropertiesCallback(true);
-				processProperties(next, {
-					properties: { ...deferredProperties, ...properties }
-				});
+				if (_idToWrapperMap.has(next.owningId)) {
```
agubler: Do not run deferred property callbacks if the widget has been destroyed

PR #430, opened 2019-06-29
Warn for functional widgets when children are not distinguishable

agubler:
**Type:** feature

**Description:**

Warn for functional widgets that are being added and removed from child lists without keys

### DIFF
--- a/src/core/vdom.ts
+++ b/src/core/vdom.ts
@@ -975,7 +975,7 @@ export function renderer(renderer: () => RenderResult): Renderer {
 					}
 				}
 			}
-			if (owningNode && isVNodeWrapper(wrapper)) {
+			if (owningNode) {
 				wrapper.owningId = owningNode.id;
 			}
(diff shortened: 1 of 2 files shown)


src/core/vdom.ts
```diff
@@ -975,7 +975,7 @@ export function renderer(renderer: () => RenderResult): Renderer {
 					}
 				}
 			}
-			if (owningNode && isVNodeWrapper(wrapper)) {
```
agubler: We should add owning node to all types of node.

src/core/vdom.ts
```diff
@@ -992,19 +992,6 @@ export function renderer(renderer: () => RenderResult): Renderer {
 		return wrappedRendered;
 	}
 
-	function findParentWNodeWrapper(currentNode: DNodeWrapper): WNodeWrapper | undefined {
```
agubler: We have a map to get parent wnode wrapper based on owning Id

PR #470, opened 2019-07-26
support assertionTemplate operations on array root

### LINKED ISSUE #434: Unable to insert into assetion template array
**Bug** 

The `.insertBefore`/`.insertAfter` methods should not throw a `TypeError: Cannot read property 'children' of undefined` error when trying to insert at the top level of an assertion template. This is caused because the found element doesn't have a parent.

Package Version: 6.0.0-alpha.8

**Code**
```typescript
const template = assertionTemplate(() => [
    <div assertion-key="first" />
    <div />
]);
template.insertAfter('~first', () => (<div />));
```

**Expected behavior:**
The element(s) should be inserted at the top level of the template

**Actual behavior:**
Error

> TypeError: Cannot read property 'children' of undefined

**Suggested Fix**
The insert operation should replace `parent.children` references with a variable that could either be the parent children or the top level array.

https://github.com/dojo/framework/blob/a41f807af9204a9d0739d1fa8e6d92d3d9e97254/src/testing/assertionTemplate.ts#L139-L159

aciccarello:
**Type:** bug

Allows insert, remove, and replace operations on nodes in a root array of an assertion template.

**Description:**

Resolves #434

### DIFF
--- a/tests/testing/unit/assertionTemplate.tsx
+++ b/tests/testing/unit/assertionTemplate.tsx
@@ -66,6 +66,25 @@ class ListWidget extends WidgetBase {
 
 const baseListAssertion = assertionTemplate(() => v('div', { classes: ['root'] }, [v('ul', [])]));
 
+class MultiRootWidget extends WidgetBase<{ after?: boolean; last?: boolean }> {
+	render() {
+		const { after, last = true } = this.properties;
+		const result = [v('div', ['first'])];
+		if (after) {
+			result.push(v('div', ['after']));
+		}
+		if (last) {
+			result.push(v('div', ['last']));
+		}
+		return result;
+	}
+}
+
+const baseMultiRootAssertion = assertionTemplate(() => [
+	v('div', { '~key': 'first' }, ['first']),
+	v('div', { '~key': 'last' }, ['last'])
+]);
+
 const tsxAssertion = assertionTemplate(() => (
 	<div classes={['root']}>
 		<h2>hello</h2>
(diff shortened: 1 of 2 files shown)


tests/testing/unit/assertionTemplate.tsx
```diff
@@ -142,6 +161,24 @@ describe('assertionTemplate', () => {
 		h.expect(childAssertion);
 	});
 
+	it('can remove a node in the root', () => {
+		const h = harness(() => w(MultiRootWidget, { last: false }));
+		const insertionAssertion = baseMultiRootAssertion.remove('~last');
+		h.expect(insertionAssertion);
+	});
+
+	it('can replacee a node', () => {
```
matt-gadd: is this `remove` too? 😄

aciccarello: You don't think I meant `replacee`? 😆

aciccarello:
@matt-gadd I've renamed that test

tests/testing/unit/assertionTemplate.tsx
```diff
@@ -142,6 +161,24 @@ describe('assertionTemplate', () => {
 		h.expect(childAssertion);
 	});
 
+	it('can remove a node in the root', () => {
+		const h = harness(() => w(MultiRootWidget, { last: false }));
+		const insertionAssertion = baseMultiRootAssertion.remove('~last');
+		h.expect(insertionAssertion);
+	});
+
+	it('can remove a node', () => {
```
agubler: @aciccarello this test seems to be duplicated and is causing intern to get upset... the exact same test is two tests above, could have been accidentally left there when merging a conflicts perhaps?

aciccarello: @agubler As I look at that more, I think that was meant to be a replace node test. Sorry. 😞

PR #475, opened 2019-07-30
Reference guide polish

sbinge:
**Type:** documentation

**Description:**

- Make top-level sections more relevant for the site nav and less
redundant given the context of the guide they're displayed in
- Remove parentheses from top-level sections names for linkability
- Align the stores guide with the latest formatting updates for display on the site


docs/en/creating-widgets/supplemental.md
```diff
@@ -1,4 +1,4 @@
-# Introduction to widgets
+# Creating widgets
```
agubler: Same title as the actual guide itself? Do you think the overall guide name should be changed?

sbinge: Think the guide name makes sense - I've changed this to something else now that isn't just duplicating the guide name.

PR #482, opened 2019-08-02
Update testing docs to use functional widgets

vansimke:
**Type:** bug / feature

* [X] All code has been formatted with [`prettier`](https://prettier.io/) as per the [readme code style guidelines](./../#code-style)

**Description:**
Updated introductory and supplemental testing guides to use functional widgets and reflect v6 module paths.

### DIFF
--- a/docs/en/testing/introduction.md
+++ b/docs/en/testing/introduction.md
@@ -1,13 +1,13 @@
 # Introduction
 
 Dojo provides a robust testing framework using `@dojo/cli-test-intern`. It allows you to efficiently test the output of your widgets and validate your expectations.
-``
-| Feature | Description |
+
+| Feature                 | Description                                                                                                                                 |
 | ----------------------- | ------------------------------------------------------------------------------------------------------------------------------------------- |
-| **Minimal API** | Simple API for testing and asserting Dojo widget's expected virtual DOM and behavior. |
-| **Unit tests** | Unit tests are tests run via node and browser to test isolated blocks of code. |
-| **Functional tests** | Functional tests are run using Selenium in the browser and test the overall functionality of the software as a user would interact with it. |
-| **Assertion templates** | Assertion Templates allow you to build expected render functions to validate the output of your widgets. |
+| **Minimal API**         | Simple API for testing and asserting Dojo widget's expected virtual DOM and behavior.                                                       |
+| **Unit tests**          | Unit tests are tests run via node and browser to test isolated blocks of code.                                                              |
+| **Functional tests**    | Functional tests are run using Selenium in the browser and test the overall functionality of the software as a user would interact with it. |
+| **Assertion templates** | Assertion Templates allow you to build expected render functions to validate the output of your widgets.                                    |
 
 # Basic usage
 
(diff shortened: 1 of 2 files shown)


docs/en/testing/introduction.md
```diff
@@ -41,24 +41,39 @@ dojo test --functional --config local
 
 ## Writing unit tests
 
--   Using Dojo's [`harness`](https://github.com/dojo/framework/tree/master/src/testing) API for unit testing widgets.
+-   Using Dojo's [`harness` API](https://github.com/dojo/framework/tree/master/src/testing#api) for unit testing widgets.
```
agubler: @vansimke what is this supposed to be linking to?

docs/en/testing/introduction.md
```diff
@@ -109,61 +124,78 @@ Assertion templates provide a way to create a base assertion that allow parts of
...
-		const h = harness(() => w(Profile, {}));
+		const h = harness(() => <Profile />);
 		// Test against my base assertion
 		h.expect(profileAssertion);
 	});
 });
 ```
 
-A value can be provided to any virtual DOM node under test using the `~key` properties defined in the assertion template. In `.tsx` this would be the `assertion-key` attribute.
+A value can be provided to any virtual DOM node under test using `assertion-key` properties defined in the assertion template. Note: when using class-based widgets, the `~key` property serves the same purpose.
```
agubler: This isn't only for class-based widgets - its usable when using the functional `w()` and `v()` APIs

docs/en/testing/supplemental.md
```diff
@@ -497,85 +496,86 @@ import assertionTemplate from '@dojo/framework/testing/assertionTemplate';
...
 
   it('renders given username correctly', () => {
-    // update the expected result with a given username
-    const namedAssertion = profileAssertion.setChildren('~welcome', [
-      'Welcome Kel Varnsen!'
-    ]);
-    const h = harness(() => w(Profile, { username: 'Kel Varnsen' }));
-    h.expect(namedAssertion);
-  });
+		// update the expected result with a given username
```
agubler: is this formatting slightly off?

docs/en/testing/introduction.md
vansimke: That links to the API documentation for the test harness in the testing README.md. That URL works for me. Do you think I should explicitly add the README.md call out -  https://github.com/dojo/framework/blob/master/src/testing/README.md#api?

agubler: We won't have the README once the documentation is done, so it will need to link to somewhere on dojo.io ideally

vansimke: Ah. That's good to know. I'll use `learn/testing/dojo-test-harness#harness-api` instead.

docs/en/testing/supplemental.md
vansimke: fixed

docs/en/testing/introduction.md
vansimke: changed wording to improve accuracy

agubler:
The CI failures are not related - am looking into it.

PR #490, opened 2019-08-12
registerCustomElement: detect child type

### LINKED ISSUE #304: Automatically detect custom element child types
**Enhancement**
The custom element descriptor currently needs to know what type of child it can take, which can be the following:
* a dojo widget
* a text node
* a dom node

With the new work around automatically deducing the custom element descriptor (https://github.com/dojo/webpack-contrib/issues/43), the child type is the only thing left that currently needs to be passed manually if it differs. It would be nice if we could auto detect this at runtime rather than explicitly having to know it up front.

mwistrand:
**Type:** feature

**Description:**

When registering custom elements, infer the `childType` from the actual child when no `childType` is explicity provided. Note that the existing behavior does not change when a `childType` is provided.

Verified locally with a demo app that contains custom elements both in `src/index.html` (rendered directly as HTML elements) and in a widgets TypeScript file (rendered with `v()`).

Resolves #304

### DIFF
--- a/src/core/registerCustomElement.ts
+++ b/src/core/registerCustomElement.ts
@@ -39,11 +39,7 @@ export function DomToWidgetWrapper(domNode: HTMLElement): any {
 }
 
 export function create(descriptor: any, WidgetConstructor: any): any {
-	const {
-		attributes = [],
-		childType = CustomElementChildType.DOJO,
-		registryFactory = () => new Registry()
-	} = descriptor;
+	const { attributes = [], childType, registryFactory = () => new Registry() } = descriptor;
 	const attributeMap: any = {};
 
 	attributes.forEach((propertyName: string) => {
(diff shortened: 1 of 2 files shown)


rorticus:
I tested this with @dojo/widgets and i'm having some issues.  The button widget has a decorator that looks like,

```ts
@customElement<ButtonProperties>({
	tag: 'dojo-button',
	childType: CustomElementChildType.TEXT,
	properties: ['disabled', 'pressed', 'popup', 'theme', 'aria', 'extraClasses', 'classes'],
	attributes: ['widgetId', 'name', 'type', 'value'],
	events: [
		'onBlur',
		'onChange',
		'onClick',
		'onFocus',
		'onInput',
		'onKeyDown',
		'onKeyPress',
		'onKeyUp',
		'onMouseDown',
		'onMouseUp',
		'onTouchCancel',
		'onTouchEnd',
		'onTouchStart'
	]
})
```

If I remove the `childType` property on that, it looks like it's assuming the child type is `DOJO` instead of `TEXT` and I get some messed up output in the custom element showcase.

image

You can see in the element inspector that the text nodes didn't get moved right.

image

mwistrand:
Closing in favor of #494

PR #495, opened 2019-08-19
Fixing focus

rorticus:
**Type:** bug

**Description:**

You currently cannot have an attribute named `focus` as this conflicts with the existing `focus` function. To solve this, we map special names like this (currently _just_ `focus`) into new names, `hasfocus` that will come through in the properties in their place.


agubler:
superseded by #499

PR #501, opened 2019-08-21
 translate docs/creating-widgets to simplified chinese

xiaohulu:
**Type:** bug / feature

**Description:**

 translate docs/creating-widgets to simplified chinese
 translate docs/middleware to simplified chinese


docs/zh-CN/creating-widgets/supplemental.md
```diff
@@ -0,0 +1,861 @@
...
+# 最佳开发实践
+
+使用 Dojo 部件时，应谨记一些重要原则，以避免在应用程序代码中引入反模式。试图以不受支持的方式使用框架可能会导致意外的行为，并在应用程序中引入难以发现的错误。
+
+## 部件属性
+
+-   部件应只能读取传入其中的属性(properties)。
+    -   如果修改了传入部件中的属性值，则不能回传给框架，以避免导致部件和框架之间出现差异。
+-   Widgets should avoid deriving further render state from their properties, and instead rely on their complete render state being provided to them.
+    -   Deriving render state can cause similar divergences between the widget and the framework as modifying received properties; the framework is not aware of the derived state, so cannot properly determine when a widget has been updated and requires invalidation and re-rendering.
```
xiaohulu: I do not understand this section.

sbinge: Hi @xiaohulu, thanks for these translations btw!

This section is essentially: don't expect Dojo to know about any derived state within a widget. Dojo only knows about the properties passed to a widget, so if any extra state is created & used for rendering, Dojo won't know when the extra state has been updated, so won't be able to re-render the widget. Widgets should be kept as simple as possible and rely on their state being passed to them.

Does that help?

xiaohulu: @sbinge Thanks. What does derived state means? Show some code that anti-pattern may help.

sbinge: Hi @xiaohulu - sorry for the delayed response, my github notifications have been playing up.

The following is a silly example of derived state:

```tsx

const factory = create().properties<{ currentValue: string }>();

function getDerivedState(value: string) {
	return `Current value is '${value}' and the current time is ${new Date().toISOString()}`;
}

export const WidgetWithDerivedState = factory(function WidgetWithDerivedState({ properties }) {
	const { currentValue } = properties();
	let derivedState = getDerivedState(currentValue);
	return (
		<div>
			<label for="dataEntry">Please enter a value:</label>
			<input
				id="dataEntry"
				value={currentValue}
				oninput={(e) => {
					derivedState = getDerivedState((e.target as HTMLInputElement).value);
				}}
			/>
			<div>{derivedState}</div>
		</div>
	);
});
```

There are two derived state anti-patterns here:

1. `getDerivedState()` uses data that changes independently of the widget's original state - in this example, it uses the current date/time.
1. the `oninput` event handler is changing the widget's state internally.

For both these anti-patterns, Dojo doesn't know that the widget's state has changed and therefore doesn't ever re-render the widget.

xiaohulu: @sbinge Thanks, it's so clear. I will translate this section.

PR #510, opened 2019-08-31
[WIP] Expose factory for typing icache to an interface

### LINKED ISSUE #509: Allow icache middleware to be typed by an interface
**Enhancement**

The `icache` middleware provides type inference when using `getOrSet` and allows a generic type to be passed to both `set` and `get`, however there is no current way to type the cache keys and their associated value types.

It would be could if `icache` could expose a factory that allowed a user to pass an interface/type and the resulting `icache` middleware would use that interface to type both the keys and the associated values of the keys. `getOrSet` would still need to support being able to pass a function that returns the associated value type or a promise that resolves to that type.

```ts
interface MyWidgetState {
    foo: {
        bar: number;
    };
}

const icache = createICacheMiddleware<MyWidgetState>();
const factory = create({ icache });

const MyWidget = factory(function MyWidget({ middleware: { icache }) {
	// returns the type `{ bar: number }`
	const resultOne = icache.getOrSet('foo', { bar: 1 });
	// errors as the value being set does not match the interface type
	icache.getOrSet('foo', { bar: 'string' });
	// returns the type `{ bar: number }`
	const resultTwo = icache.getOrSet('foo', () => ({ bar: 1 }));
	// errors as the value being set does not match the interface type
	icache.getOrSet('foo', () => ({ bar: 'string' }));
	// returns the type `{ bar: number } | undefined`
	const resultThree = icache.getOrSet('foo', async () => ({ bar: 1 }));
	// errors as the value being set does not match the interface type
	icache.getOrSet('foo', async () => ({ bar: 'string' }));
	// Returns the type `{ bar: number } | undefined`
	const resultFour = icache.get('foo');
	// no errors setting the correct type
	icache.set('foo', { bar: 1 });
	// errors if the type does not match the interface
	icache.set('foo', 'string');
});
```

It is important to maintain the current typing for users that want to user the current icache inferred and generic typings.

agubler:
**Type:** bug

**Description:**

Adds factory for typing the `icache` middleware to an interface. Resolves typings as described in the associated issue but maintains the existing typings when using the default `icache`

Resolves #509


docs/en/middleware/supplemental.md
```diff
@@ -201,6 +201,59 @@ import icache from '@dojo/framework/core/middleware/icache';
 -   `clear()`
     -   Clears all values currently stored in the widget's local cache.
 
+`icache` can be typed in two different ways, the first is using generics that enables the return type to be specified at the call-site and for `getOrSet` the return type can be inferred from the value type. If the `value` for `getOrSet` is a function then the type will be inferred from the functions return type.
```
dylans:
```suggestion
`icache` can be typed in two different ways. One approach uses generics to enable the return type to get specified at the call-site, and for `getOrSet`, the return type can get inferred from the value type. If the `value` for `getOrSet` is a function then the type will get inferred from the functions return type.
```

docs/en/middleware/supplemental.md
```diff
@@ -201,6 +201,59 @@ import icache from '@dojo/framework/core/middleware/icache';
...
+		const response = await fetch('url');
+		const body: FetchResult = await response.json();
+		return body;
+	});
+
+	return <div>{results}</div>;
+});
+```
+
+This however doesn't give any typing for the cache keys, the preferred way to type `icache` is to create a pre-typed middleware using `createICacheMiddleware`. This allows for passing an interface which will create an `icache` middleware typed specifically for the passed interface and provides type safety for the cache keys.
```
dylans:
```suggestion
However this approach doesn't provide any typing for the cache keys. The preferred way to type `icache` is to create a pre-typed middleware using `createICacheMiddleware`. This allows for passing an interface which will create an `icache` middleware typed specifically for the passed interface and provides type safety for the cache keys.
```

PR #520, opened 2019-09-09
Add 'isExact' option to active link

### LINKED ISSUE #492: Active Link should allow specifying whether a link required an exact match
**Enhancement** 

Right now an active link will always apply active classes for child routes if the parameters are not different. It'd be nice to be able to specify that a link requires an exact match to be considered active.
Package Version: 6.0.0-alpha.16

maier49:
**Type:** feature

**Description:**
Adds an `isExact` option to the `ActiveLink` properties that if true will only add active classes if the matched outlet's `isExact` function returns `true`.
Resolves #492

### DIFF
--- a/src/routing/ActiveLink.ts
+++ b/src/routing/ActiveLink.ts
@@ -8,6 +8,7 @@ import Router from './Router';
 
 export interface ActiveLinkProperties extends LinkProperties {
 	activeClasses: SupportedClassName[];
+	isExact?: boolean;
 }
 
 function paramsEqual(linkParams: any = {}, contextParams: any = {}) {
@@ -22,7 +23,7 @@ export const ActiveLink = factory(function ActiveLink({
 	children
 }) {
 	const { to, routerKey = 'router', params } = properties();
-	let { activeClasses, classes = [], ...props } = properties();
+	let { activeClasses, isExact, classes = [], ...props } = properties();
 
 	diffProperty('to', (current: ActiveLinkProperties, next: ActiveLinkProperties) => {
 		if (current.to !== next.to) {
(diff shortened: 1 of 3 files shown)


src/routing/ActiveLink.ts
```diff
@@ -8,6 +8,7 @@ import Router from './Router';
 
 export interface ActiveLinkProperties extends LinkProperties {
 	activeClasses: SupportedClassName[];
+	isExact?: boolean;
```
agubler: Is having an `isExact` property a breaking change by default? As if it is not passed then it will fall back to a partial match? Would it be better to have `isPartial`, with the default being `false`/`undefined` and therefore requiring an exact match.

Another thought, do you think there would ever be a need to want to match specific params partially? In dojo/site we partially match for a specific param depending on the link being rendered, if we just have a boolean option of `isPartial` or `isExact` it would actually consider every link active even though they are each for a specific param?

agubler: This is the example of dojo/site - https://github.com/dojo/site/blob/3be05e769914981c13dcc377d720367d8e0d5f9c/src/learn/Learn.tsx#L46

maier49: I don't see how `isExact` is a breaking change. By default it's not specified, and if it's false the behavior is the same as what's current: https://github.com/dojo/framework/pull/520/files#diff-3e02e2cbd03b930052645ddb1510bc2dR62

Do you mean it's breaking if it's `true`, because you can't specify `isExact` on an `ActiveLink` currently so how would that break?

I can certainly look into adding something to match specific params. But, could that be separate from this? As I understand it all this change is doing is giving an option so that you can opt in to only count links as active if the context that would get passed to your link has `isExact()` returning true. If you don't specify the property everything behaves as it does now and this could coexist with any other changes to partially match params, as currently any partial matches would be excluded regardless of this change.

agubler: You're absolutely right, sorry I was thinking about matching params "exactly" for a route which is a different (but useful) change to this.

PR #523, opened 2019-09-10
Ensure middleware gets destroyed from deepest to shallowest

### LINKED ISSUE #518: destroy middleware FIFO
**Bug**

The `destroy` middleware is useful for making sure that memory is freed when a widget/middleware using middleware is destroyed.

However, the `destroy` middleware seems to operate on a FIFO basis. This means that the dependencies of a middleware or widget are destroyed before the dependants. This makes it impossible to use middlewares during the destroy process.

I made a CodeSandbox example that shows what happens when a `dependant` middleware and a `dependency` middleware both use `destroy`

https://codesandbox.io/embed/dojo-codesandbox-template-p9jg8

The outcome of this behavior is that dependencies are not available during cleanup. I ran into this issue when I was using `cache` to hold a collection of `Handle` so their `destroy()` method could be called on `destroy`. Unfortunately the `cache` middleware had already been `destroyed` by the time my middleware could be cleaned up making the handles inaccessible.

Instead we should change the behavior of `destroy()` to be FILO so widgets will have their dependencies available for any special cleanup, then dependencies, and dependencies of dependencies, and so on.

Package Version:  6

image

agubler:
**Type:** bug

**Description:**

The destroy middleware is useful for ensuring widgets can deal with any references when they are removed from the tree. However currently, destroy is called in the order that the middleware dependencies of a widget call destroy. This is generally top down as `destroy` is usually called at the point of a middlewares initialisation, but this is not a guarantee meaning that not only is it generally called in the wrong order but also is not deterministic.

This change records the dependency order of all middleware for a widget and when there have been destroy functions registered ensures that they called in the right order from the deepest to shallowest (first in, last out).

Resolves #518

### DIFF
--- a/src/core/vdom.ts
+++ b/src/core/vdom.ts
@@ -73,6 +73,7 @@ export interface WidgetMeta {
 	dirty: boolean;
 	invalidator: () => void;
 	middleware?: any;
+	middlewareIds: string[];
 	registryHandler?: RegistryHandler;
 	registry: Registry;
 	properties: any;
@@ -698,8 +699,11 @@ function addNodeToMap(id: string, key: string | number, node: HTMLElement) {
 	}
 }
 
-function destroyHandles(destroyMap: Map<string, () => void>) {
-	destroyMap.forEach((destroy) => destroy());
+function destroyHandles(destroyMap: Map<string, () => void>, middlewareIds: string[]) {
+	for (let i = 0; i < middlewareIds.length; i++) {
+		const destroy = destroyMap.get(middlewareIds[i]);
+		destroy && destroy();
+	}
 	destroyMap.clear();
 }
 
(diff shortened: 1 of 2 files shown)


src/core/vdom.ts
```diff
@@ -698,8 +699,11 @@ function addNodeToMap(id: string, key: string | number, node: HTMLElement) {
 	}
 }
 
-function destroyHandles(destroyMap: Map<string, () => void>) {
-	destroyMap.forEach((destroy) => destroy());
+function destroyHandles(destroyMap: Map<string, () => void>, middlewareIds: string[]) {
```
devpaul: (optional) since adding the `middlewareIds` this method has become less generic. We could refactor it as `destroyHandles(meta: WidgetMeta)` and move the `destroyMap` existence check internal.

agubler: @devpaul updated :+1: thanks

PR #536, opened 2019-09-17
Middlware mocking documentation

### LINKED ISSUE #519: Add site content for how to mock in dojo
**Enhancement** 

We've had a number of people on discord ask how to mock a variety of things in tests from `fetch` to middlewares such as ResizeObserver. It would be good if we had content describing how to achieve those things.

mocking shims: https://github.com/dojo/framework/pull/295

mock middleware: https://github.com/dojo/framework/tree/master/src/testing/mocks/middleware

jellison:
**Type:** feature

**Description:**

Introduces an icache middleware mock as well as documentation for using it and creating custom middleware mocks.

Resolves #519

### DIFF
--- a/src/testing/mocks/middleware/icache.ts
+++ b/src/testing/mocks/middleware/icache.ts
@@ -1,0 +1,73 @@
+/* tslint:disable:interface-name */
+import { create, invalidator } from '../../../core/vdom';
+import { MiddlewareResult } from '../../../core/interfaces';
+import Map from '../../../shim/Map';
+
+export function createICacheMock() {
+	const map = new Map<any, any>();
+	const factory = create({ invalidator });
+
+	const mockICacheFactory = factory(({ middleware: { invalidator } }) => {
+		return {
+			getOrSet<T = any>(key: any, value: any): T | undefined {
+				if (map.has(key)) {
+					return this.get(key);
+				} else {
+					this.set(key, value);
+					return undefined;
+				}
+			},
+			get<T = any>(key: any): T | undefined {
+				if (map.has(key)) {
+					const mapValue = map.get(key);
+					if (typeof mapValue === 'function') {
+						return undefined;
+					} else {
+						return mapValue as T;
+					}
+				}
+				return undefined;
+			},
+			set(key: any, value: any): void {
+				if (typeof value === 'function') {
+					value = value();
+					if (value.then && typeof value.then === 'function') {
+						value.then((result: any) => {
+							map.set(key, result);
+							invalidator();
+						});
+					}
+					map.set(key, value);
+				} else {
+					map.set(key, value);
+					invalidator();
+				}
+			},
+			clear(): void {}
+		};
+	});
+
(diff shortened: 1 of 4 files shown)


src/testing/mocks/middleware/icache.ts
```diff
@@ -0,0 +1,73 @@
+/* tslint:disable:interface-name */
+import { create, invalidator } from '../../../core/vdom';
+import { MiddlewareResult } from '../../../core/interfaces';
+import Map from '../../../shim/Map';
+
+export function createICacheMock() {
+	const map = new Map<any, any>();
+	const factory = create({ invalidator });
+
+	const mockICacheFactory = factory(({ middleware: { invalidator } }) => {
```
agubler: Do we need to recreate the `icache` middleware logic in the mock? It would be good if we could pass everything off to the real `icache` and only intercept the `set` call to get the promise (if it is one)?

There are examples in some of the other mock middleware that effectively do similar things, like the `resize` mock. Something like:

```ts
	const factory = create({ cache, invalidator });

	const mockICacheFactory = factory(({ id, middleware, properties, children }) => {
		const { callback } = icache();
		invalidate = middleware.invalidator;
		const icacheMiddleware = callback({
			id,
			middleware: { invalidator: middleware.invalidator, cache: middleware.cache },
			properties,
			children
		});
        return {
            getOrSet(key: any, value: any) {
                return icacheMiddleware.getOrSet(key, value);
            }
            get(key: any) {
               icacheMiddleware.get(key);
            }
            set(key: any, value: any) {
                // do things in here to get the promise to make it available to the mock
            }
            clear() {
               icacheMiddleware.clear();
            }
        }
	});
```

jellison: Yep, seems possible.

agubler:
Hey @jellison I just noticed that this is raised from a branch on dojo/framework, we work from forks for PRs, do you think you could move it to a fork?

jellison:
> Hey @jellison I just noticed that this is raised from a branch on dojo/framework, we work from forks for PRs, do you think you could move it to a fork?

Yeah. I'll have to create a new PR though.

jellison:
Replaced with PR #537

src/testing/mocks/middleware/icache.ts
matt-gadd: yeah that sounds good, I think we can even just monkey patch the icache that comes back so we'd only have to implement `set` and not deal with the rest? something like*:

```tsx
const factory = create({ cache, invalidator });
const mockICacheFactory = factory(({ id, middleware, properties, children }) => {
  const { callback } = icache();
  invalidate = middleware.invalidator;
  const icacheMiddleware = callback({ id, middleware: { invalidator: middleware.invalidator, cache: middleware.cache }, properties, children }); 
  const setter = icacheMiddleware.set;
  icaceMiddleware.set = (...args) => {
    // whatever
    return setter(...args);
  }
  return icacheMiddleware;
});
```
* t&c's apply, totally pseudo code 😄

PR #537, opened 2019-09-17
Middleware mocking documentation

### LINKED ISSUE #519: Add site content for how to mock in dojo

jellison:
**Type:** feature

**Description:**

Introduces an icache middleware mock as well as documentation for using it and creating custom middleware mocks.

Resolves #519

### DIFF
--- a/src/testing/mocks/middleware/icache.ts
+++ b/src/testing/mocks/middleware/icache.ts
@@ -1,4 +1,3 @@
-/* tslint:disable:interface-name */
 import { create, invalidator } from '../../../core/vdom';
 import { MiddlewareResult } from '../../../core/interfaces';
 import { cache } from '../../../core/middleware/cache';
@@ -21,13 +20,14 @@ export function createICacheMock() {
 		icacheMiddleware.set = (key: any, value: any) => {
 			if (typeof value === 'function') {
 				value = value();
-				map.set(key, value);
 				if (value && typeof value.then === 'function') {
+					map.set(key, value);
 					setter(key, () => value);
-				} else {
-					setter(key, value);
+					return;
 				}
 			}
+
+			setter(key, value);
 		};
 
 		return icacheMiddleware;
@@ -38,13 +38,7 @@ export function createICacheMock() {
 	function mockCache(key?: string): Promise<any> | MiddlewareResult<any, any, any> {
 		if (key) {
 			if (map.has(key)) {
-				const mapValue = map.get(key);
-
-				if (mapValue.then && typeof mapValue.then === 'function') {
-					return mapValue;
-				} else {
-					return Promise.resolve(mapValue);
-				}
+				return map.get(key);
 			} else {
 				return Promise.resolve(undefined);
 			}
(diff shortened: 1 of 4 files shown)


jellison:
@agubler 51015c12db24f32311cf6eb6dbda6f0a3d32e582 addresses your concern about duplicating the cache implementation.

src/testing/mocks/middleware/icache.ts
```diff
@@ -0,0 +1,59 @@
+/* tslint:disable:interface-name */
```
agubler: Do we need this?

src/testing/mocks/middleware/icache.ts
```diff
@@ -0,0 +1,59 @@
...
+			middleware: { invalidator: middleware.invalidator, cache: middleware.cache },
+			properties,
+			children
+		});
+		const setter = icacheMiddleware.set;
+
+		icacheMiddleware.set = (key: any, value: any) => {
+			if (typeof value === 'function') {
+				value = value();
+				map.set(key, value);
```
agubler: Do you need to set this in the exposed map if it's not a promise?

src/testing/mocks/middleware/icache.ts
```diff
@@ -0,0 +1,59 @@
...
+		const icacheMiddleware = callback({
+			id,
+			middleware: { invalidator: middleware.invalidator, cache: middleware.cache },
+			properties,
+			children
+		});
+		const setter = icacheMiddleware.set;
+
+		icacheMiddleware.set = (key: any, value: any) => {
+			if (typeof value === 'function') {
```
agubler: I think we need to call the original setter with the value if it's not a function too.

src/testing/mocks/middleware/icache.ts
jellison: Are you thinking the mock would fall back to the middleware if the `map` did not contain a key? It doesn't now, but if it did, then no; the `map` would only need to track the promises.

src/testing/mocks/middleware/icache.ts
jellison: Definitely.

src/testing/mocks/middleware/icache.ts
agubler: I didn't know the reason why it would need to expose values that weren't a promise, the only reason we need to expose the promise is to await the resolution?

jellison: I'm fine with that. Users can always expand the idea with their own mock if they have specific needs.

jellison: Resolved with c45e07aab8ab50d633b5f548a2c5bede22ef2c48

PR #544, opened 2019-09-29
Typed children for function-based widgets

### LINKED ISSUE #43: Support typed children in TSX
@matt-gadd commented on [Wed Sep 13 2017](https://github.com/dojo/widget-core/issues/675)

**Enhancement** 

TypeScript now supports typed children in TSX, we need to update our JSX namespace to leverage it by adding the `ElementChildrenAttribute`.

---

@matt-gadd commented on [Tue Jun 05 2018](https://github.com/dojo/widget-core/issues/675#issuecomment-330181036)

Okay this doesn't currently work due to `ElementChildrenAttribute` looking on the `ElementAttributesProperty`, which in our case is `properties`. See https://github.com/Microsoft/TypeScript/issues/18357

### LINKED ISSUE #528: Support named children
**Enhancement** 
Up to now when when we've been using render props in widgets we've used a by convention `renderer` property. This kind of works for a single renderer, but not for many and because this is only by convention we can't for certain know what property is a render prop. In 7 we'd like to propose a change to our children behaviour to support named renderers, it would look something like:

```tsx
<SplitPane>
{
  {
    left() => <div>Left Item</div>,
    right() => <div>Right Item</div>
  }
}
</SplitPane>
```

or

```tsx
<SplitPane>
{
  () => ({
    left: <div>Left Item</div>,
    right: <div>Right Item</div>
  })
}
</SplitPane>
```

As part of this we might need to move `children` to the properties interface to allow this to be inferred by TSX.

With this in place we should be able to write a better harness to deal with render props.

agubler:
**Type:** feature

**Description:**

Provides extended typing of children for function-based widgets, supported by all widget authoring patterns. When children types are defined it makes them mandatory for widget usage.

```tsx
import { create, tsx } from '@dojo/framework/core/vdom';

const factory = create().children<(value: string) => RenderResult>();

const Foo = factory(function Foo({ children }) {
    const [childRender] = children();
    return childRender('text');
});

// usages

// using TSX
<Foo>{(value) => <div>value</div>}</Foo>

// Using the widget function
Foo({}, [(value) => <div>value</div>]); 

// Using the `w()` pragma
w(Foo, {}, [(value) => <div>value</div>]);
```

Typed children provide a first class replacement for widgets that use render props, for example `Outlet`.

```tsx
<Outlet id="outlet" renderer={(matchDetails) => {
    if (matchDetails.isExact()) {
        return <Home />;
    }
    return null;
}}/>

<Outlet id="outlet">{(matchDetails) => {
    if (matchDetails.isExact()) {
        return <Home />;
    }
    return null;
}}</Outlet>
```

Resolves #528 
Resolves #43

### DIFF
--- a/src/core/interfaces.d.ts
+++ b/src/core/interfaces.d.ts
@@ -394,7 +394,7 @@ export interface Callback<Props, Middleware, ReturnValue> {
 			id: string;
 			middleware: MiddlewareApiMap<Middleware>;
 			properties: () => Props & { children?: never };
-			children: () => Props extends { children: any } ? Props['children'] : DNode[];
+			children: () => (Props extends { children: any } ? Props['children'] : DNode[]) | undefined;
 		}
 	): ReturnValue;
 }
--- a/src/core/interfaces.d.ts
+++ b/src/core/interfaces.d.ts
@@ -393,8 +393,8 @@ export interface Callback<Props, Middleware, ReturnValue> {
 		options: {
 			id: string;
 			middleware: MiddlewareApiMap<Middleware>;
-			properties: () => Props & { children?: never };
-			children: () => (Props extends { children: any } ? Props['children'] : DNode[]) | undefined;
+			properties: () => Props & { __children__?: never };
+			children: () => (Props extends { __children__: any } ? Props['__children__'] : DNode[]) | undefined;
 		}
 	): ReturnValue;
 }
@@ -412,8 +412,8 @@ export interface MiddlewareResultFactory<Props, Middleware, ReturnValue> {
 
 export interface WNodeFactory<W extends WidgetBaseTypes> {
 	(
-		properties: WithOptional<W['properties'], 'children'>,
-		children?: W['properties'] extends { children?: any } ? W['properties']['children'] : W['children']
+		properties: WithOptional<W['properties'], '__children__'>,
+		children?: W['properties'] extends { __children__?: any } ? W['properties']['__children__'] : W['children']
 	): WNode<W>;
 }
 
--- a/src/core/interfaces.d.ts
+++ b/src/core/interfaces.d.ts
@@ -4,6 +4,10 @@ import Map from '../shim/Map';
 import WeakMap from '../shim/WeakMap';
 import { RegistryHandler } from './RegistryHandler';
 
+export type OmitProp<T, K extends keyof any> = Pick<T, Exclude<keyof T, K>>;
+export type Omit<T, K extends keyof T> = Pick<T, Exclude<keyof T, K>>;
+export type WithOptional<T, K extends keyof T> = Omit<T, K> & Partial<Pick<T, K>>;
+
 /**
  * Generic constructor type
  */
@@ -395,7 +399,7 @@ export interface Callback<Props, Middleware, ReturnValue> {
 	): ReturnValue;
 }
 
-export interface MiddlewareResult<Props  extends { children?: {} }, Middleware, ReturnValue> {
+export interface MiddlewareResult<Props extends { children?: {} }, Middleware, ReturnValue> {
 	api: ReturnValue;
 	properties: Props;
 	callback: Callback<Props, Middleware, ReturnValue>;
@@ -407,7 +411,10 @@ export interface MiddlewareResultFactory<Props, Middleware, ReturnValue> {
 }
 
 export interface WNodeFactory<W extends WidgetBaseTypes> {
-	(properties: W['properties'], children?: W['children']): WNode<W>;
+	(
+		properties: WithOptional<W['properties'], 'children'>,
+		children?: W['properties'] extends { children?: any } ? W['properties']['children'] : W['children']
+	): WNode<W>;
 }
 
 export type UnionToIntersection<U> = (U extends any ? (k: U) => void : never) extends ((k: infer I) => void)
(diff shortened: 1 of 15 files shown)


src/core/interfaces.d.ts
```diff
@@ -4,6 +4,10 @@ import Map from '../shim/Map';
 import WeakMap from '../shim/WeakMap';
 import { RegistryHandler } from './RegistryHandler';
 
+export type OmitProp<T, K extends keyof any> = Pick<T, Exclude<keyof T, K>>;
+export type Omit<T, K extends keyof T> = Pick<T, Exclude<keyof T, K>>;
```
matt-gadd: will this conflict with the global `Omit` type added in 3.5? I presume not but worth a check

src/core/vdom.ts
```diff
@@ -323,8 +332,12 @@ export function w<W extends WidgetBaseTypes>(
 		| LazyDefine<W>
 		| Callback<any, any, RenderResult>,
 	properties: W['properties'],
-	children?: W['children']
+	children?: any
 ): WNode<W> {
+	if ((properties as any).children) {
```
matt-gadd: is this correct now? would it not be `__children__`?

tests/core/unit/vdom.ts
```diff
@@ -3430,6 +3430,42 @@ jsdomDescribe('vdom', () => {
 				assert.isTrue(consoleWarnStub.calledOnce);
 			});
 
+			it('typed children', () => {
```
matt-gadd: do we need an example text with `w()` also?

tests/testing/unit/support/selector.ts
```diff
@@ -70,7 +70,7 @@ describe('selector', () => {
 			assert.deepEqual(adapter.getChildren(v('div', {}, [w(WidgetBase, {})])), [w(WidgetBase, {})]);
 			assert.deepEqual(adapter.getChildren(w(WidgetBase, {}, [v('span')])), [v('span')]);
 			assert.deepEqual(adapter.getChildren(w(WidgetBase, {}, [w(WidgetBase, {})])), [w(WidgetBase, {})]);
-			assert.deepEqual(adapter.getChildren(w(WidgetBase, {})), []);
+			assert.deepEqual(adapter.getChildren(w(WidgetBase, {})), undefined);
```
matt-gadd: how come this had to change?

agubler: Because by default children are not set to an empty array when using `w` any more.

tests/core/unit/vdom.ts
agubler: I can add that

src/core/vdom.ts
agubler: Nice catch

src/core/interfaces.d.ts
agubler: It shouldn't do, I did a quick check and there no complaints about `Omit`

tests/testing/unit/support/selector.ts
matt-gadd: but this is the assertion for `getChildren` no? which is both typed as an array and returns an empty array if not set?

agubler: If it is a WNode or VNode then it just returns the children from the element.

tests/core/unit/vdom.ts
agubler: Added.

src/core/vdom.ts
agubler: Fixed and a couple of other misses too.

tests/testing/unit/support/selector.ts
matt-gadd: Ah okay, sorry I missed that this was in the selector, and not the assertion template. This makes me concerned about these types/values a little bit though. 

I think any kind of `children` interface should be consistently returning either an empty array `[]` for no children, an array of one `['a']` for one item, and an array of many `['a','b','c']` for many items. This means the end user can consistently work on an array without guards for api's such as this, and would match other child like api's such as `NodeList`.

I'm not sure what the ramifications are on types or whether it's possible with `TSX`, but it'd seem more consistent to make the user define `create().children<[ whatever ]>()` over `create().children<whatever>()` if they need to define a single item.

agubler: I think that's fair point about always having children and it being an array (that can be defaulted). I've made some changes (WIP atm as need to put back the defaulting of children) that added support for children to be mandatory and able to be specified singularly but are consumed by the widget author as the first item in an array.

src/testing/mocks/middleware/intersection.ts
```diff
@@ -39,12 +39,12 @@ export function createIntersectionMock() {
 		}
 	};
 
-	function mockIntersection(): MiddlewareResult<any, any, any>;
+	function mockIntersection(): MiddlewareResult<any, any, any, any>;
```
matt-gadd: can we add a convenience type for this? it's used quite a lot in the mocks, and presumably would be also for a user writinga mock?

agubler: I added a `DefaultMiddlewareResult` interface that represents `MiddlewareResult<any, any, any, any>`

matt-gadd:
this is awesome 👍 took some wrangling in the end but worth it.

PR #545, opened 2019-09-30
Adding drag meta

### LINKED ISSUE #514: Drag Middleware
**Enhancement**
Create a middleware that provides feature parity with the existing `meta/Drag`.

rorticus:
**Type:** feature

**Description:**

Converting `Drag` meta to `drag` middleware. This is basically a direct port of the drag meta. I tried a few ways to be clever but having a single copy of the event listeners, like this has, seemed like a better way to go.

Resolves #514

### DIFF
--- a/src/core/middleware/drag.ts
+++ b/src/core/middleware/drag.ts
@@ -1,0 +1,276 @@
+import { deepAssign } from '../util';
+import global from '../../shim/global';
+import { assign } from '../../shim/object';
+import WeakMap from '../../shim/WeakMap';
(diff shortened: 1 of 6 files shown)


agubler:
@rorticus Do you think we should be using a class in the functional middleware? Could this not be done all within the middleware?

rorticus:
@agubler I can poke around with that some more if you like. I originally went that direction but ran into issues because I wanted a single global event listener, and the middleware is created for each widget. I can take another shot at that though, basically moving a lot of the class stuff up to the module scope?

PR #564, opened 2019-10-22
Provide live binding for properties and children with function-based widgets

### LINKED ISSUE #563: Binding issue with properties and functional widgets
**Bug** 

It looks like `properties()` calls are binding to an old version of the method and not the latest version.  `properties()` should always return the latest copy of the properties.

Package Version: 6.0.1

**Code**

https://codesandbox.io/s/dojo-codesandbox-template-0wfis

In this example, click on "Increment value", then "Click me"

**Expected behavior:**

The value "11" would be printed to the console (the latest properties)

**Actual behavior:**

The value "1" is printed to the console (the original properties)

agubler:
**Type:** bug

**Description:**

The `properties()` function should provide a binding to the latest live property values, so when `properties()` is called in a callback it should provide the latest property value.

Resolves #563


src/core/vdom.ts
```diff
@@ -1790,6 +1790,27 @@ export function renderer(renderer: () => RenderResult): Renderer {
...
+					return widgetMeta.properties;
+				}
+				return {};
+			},
+			children: () => {
+				const widgetMeta = widgetMetaMap.get(id);
+				if (widgetMeta) {
+					return widgetMeta.children;
+				}
+				return {};
```
rorticus: Should this be `[]`?

agubler: Yes. Ooops.

PR #593, opened 2019-11-29
Type i18n format function using the passed bundle message keys

### LINKED ISSUE #591: i18n middleware format function could be typed by message keys
**Enhancement**

The format function in the i18n middleware is currently typed as `(key: string, options?: any)`:

https://github.com/dojo/framework/blob/master/src/core/middleware/i18n.ts#L94

It would be useful to be able to ensure that the `key` argument was a type of the key of the messages in the bundle to prevent passing invalid key property.

Package Version: 7.0.0-pre

**Code**

Assume an .nls file as such:

```typescript
const messages = {
     foosAvailable: '{num} are available'
};

export default { messages };
```

A widget with this behaviour in the `render`

```typescript
const { messages, format } = i18n.localize(bundle);
// Later on
format('barsAvailable', { num: bars.length })}
```

**Expected behavior:**

A type error occurs when using a no message bundle key.

**Actual behavior:**

There is no type error.

JamesLMilner:

**Description:**

This PR prevents i18n middleware users from passing keys that aren't part of the bundle messages to the `format` function.

i18nformat

Resolves #591

### DIFF
--- a/src/core/interfaces.d.ts
+++ b/src/core/interfaces.d.ts
@@ -680,7 +680,7 @@ export type LocalizedMessages<T extends Messages> = {
 	 * @return
 	 * The formatted string.
 	 */
-	format(key: string, options?: any): string;
+	format(key: keyof T, options?: any): string;
 
 	/**
 	 * The localized messages if available, or either the default messages or a blank bundle depending on the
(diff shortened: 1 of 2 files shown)


agubler:
@JamesLMilner can you raise this change from your fork please?

PR #620, opened 2019-12-21
WIP: Set document title on route change

### LINKED ISSUE #601: Router should be able to change the document.title
It would be nice if the router and routing config would support changing the title on page visited.

agubler:
**Type:** feature

**Description:**

Adds support for setting the `document.title` on route change. This can be done either statically in the route configuration, a callback passed to the router to dynamically set the title or a combination of the two.

#### Statically:

> routes.ts

```ts
export default [
    {
        path: 'foo/{param}?{queryParam}'
        outlet: 'foo',
        title: 'Foo Title'
    }
];
```

#### Dynamically on Router initialisation using the `setDocumentTitle` callback (using the injector registration as an example, but the same can be passed when instantiating a new `Router` directly):

```ts
registerRouterInjector(routes, registry, {
    setDocumentTitle: ({ title, outlet, params, queryParams }) => {
        if (outlet === 'dynamic-title-outlet') {
            return `${outlet}-${params.id}-${queryParams.bar}`;
        }
        return title;
    }
});
```

Currently the options interface for the `setDocumentTitle` look like:

```ts
export interface DocumentTitleOptions {
	title?: string;
	outlet: string;
	params: Params;
	queryParams: Params;
}
```

Resolves #601

### DIFF
--- a/src/routing/Router.ts
+++ b/src/routing/Router.ts
@@ -1,5 +1,15 @@
+import global from '../shim/global';
 import Evented from '../core/Evented';
-import { RouteConfig, History, OutletContext, Params, RouterInterface, Route, RouterOptions } from './interfaces';
+import {
+	RouteConfig,
+	History,
+	OutletContext,
+	Params,
+	RouterInterface,
+	Route,
+	RouterOptions,
+	MatchType
+} from './interfaces';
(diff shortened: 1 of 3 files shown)


agubler:
superseded by https://github.com/dojo/framework/pull/725

PR #631, opened 2020-01-03
Add validity middleware mock and tests

### LINKED ISSUE #630: Add a validity middleware mock
**Enhancement** 

Package Version: 7.0.0

**Code**

Because this has the same API as the node middleware (a `get` method) we could just re-export that. This way, the same endpoint can be used but it just seems odd to pull in the node mock to use here.

```typescript
export { createNodeMock as createValidityMock, createNodeMock as default } from './node';
```

nicknisi:
**Type:** feature

**Description:**

This adds a validity middleware mock and tests for it.

```tsx
const App = factory(({ middleware: { validity } }) => {
	const { valid, message } = validity.get('test', 'test value');
	return !valid ? <div>{message}</div> : <div>valid</div>;
});

validityMock('test', { valid: false, message: 'test message' });
let h = harness(() => <App />, { middleware: [[validity, validityMock]] });
h.expect(() => <div>test message</div>);
```

Resolves #630

### DIFF
--- a/src/testing/mocks/middleware/validity.ts
+++ b/src/testing/mocks/middleware/validity.ts
@@ -1,0 +1,39 @@
+import { create, invalidator } from '../../../core/vdom';
+import { DefaultMiddlewareResult } from '../../../core/interfaces';
+
+interface ValidityResult {
+	valid?: boolean;
+	message?: string;
+}
+
+export function createValidityMock() {
+	const values: { [key: string]: ValidityResult } = {};
+	let invalidate: () => void | undefined;
+
+	const factory = create({ invalidator });
+
+	const mockValidityFactory = factory(({ middleware: { invalidator } }) => {
+		invalidate = invalidator;
+		return {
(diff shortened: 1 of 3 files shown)


src/testing/mocks/middleware/validity.ts
```diff
@@ -0,0 +1,39 @@
...
+	const values: { [key: string]: ValidityResult } = {};
+	let invalidate: () => void | undefined;
+
+	const factory = create({ invalidator });
+
+	const mockValidityFactory = factory(({ middleware: { invalidator } }) => {
+		invalidate = invalidator;
+		return {
+			get(key: string | number, value: string) {
+				value;
```
matt-gadd: is this meant to be here?

nicknisi: Oops, no I meant to remove that. I originally added it because `value` is unused and I was quieting TypeScript. Renaming the argument to `_value` to quiet TypeScript instead.

PR #633, opened 2020-01-06
Use MessageFormat.js; deprecate Globalize wrappers

### LINKED ISSUE #250: Remove Globalize.js in favor of `Intl` (and `MessageFormat`)
**Enhancement**

Dojo currently uses [Globalize.js](https://github.com/globalizejs/globalize) for the bulk of its internationalization/localization functionality, but at this point `Intl` is well-supported and robust enough that it can handle most use cases. For supported browsers that lack `Intl` support, there are shims available ([`Intl`](https://github.com/andyearnshaw/Intl.js) and [`Intl.RelativeFormat`](https://github.com/yahoo/intl-relativeformat)). With that in mind, we can remove Globalize.js, use [MessageFormat](https://messageformat.github.io/messageformat/) internally for managing ICU message formatting, and recommend the native `Intl` for anything else.

This would be a breaking change, and applications that use functions like `formatCurrency` would need to update to their native equivalents like `Intl.NumberFormat`. Perhaps `@dojo/cli-upgrade-app` could help mitigate any manual updates

With regard to the build, conditional polyfill loading should help reduce the bundle size, but it would also be possible to pre-compile any messages and therefore eliminate most of the bulk from the MessageFormat library as well. However, doing so will be difficult without a standard means of differentiating message bundles from other files. Just as Dojo uses `m.css` to distinguish CSS module files from other CSS files, perhaps a similar approach could be used for localized message files. The build would then pre-compile the messages in these files and `MessageFormat` could be ignored entirely. More discussion is needed here.

mwistrand:
**Type:** feature

**Description:**

Resolves #250 

Add [MessageFormat](https://messageformat.github.io/messageformat/) for ICU message formatting, and add deprecation warnings to the Globalize.js wrapper functions in `i18n/date.ts`, `i18n/number.ts`, and `i18n/unit.ts` with the intention of removing Globalize.js and CLDR.js as `@dojo/framework` dependencies in a future version.

Verified with the [World Clock and TODO MVC Kitchensink](https://github.com/mwistrand/examples/tree/feat/deprecate-globalize) example applications.

### DIFF
--- a/docs/en/i18n/supplemental.md
+++ b/docs/en/i18n/supplemental.md
@@ -705,7 +705,9 @@ i18n(bundle, 'en').then((messages) => {
 });
 ```
 
-## Date and number formatting.
+## Date, number, and unit formatting. (Deprecated)
+
+**Note**: This feature has been deprecated in favor of using [`Intl`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl) or the relevant [`Globalize.js`](https://github.com/globalizejs/globalize) methods directly.
 
 **Note**: This feature requires appropriate [CLDR data](#loading-cldr-data) to have been loaded into the application.
 
--- a/docs/en/i18n/supplemental.md
+++ b/docs/en/i18n/supplemental.md
@@ -472,11 +472,6 @@ loadCldrData({
 
 Dojo's `i18n` module requires the following CLDR data for each particular formatting feature:
 
-For [ICU message formatting](#icu-message-formatting):
-
--   `supplemental/likelySubtags`
--   `supplemental/plurals`
-
 For [date/time formatting](#date-and-number-formatting):
 
 -   `main/{locale}/ca-gregorian`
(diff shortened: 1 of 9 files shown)


matt-gadd:
I still have a few issues regarding how this is all going to tie together (it might be best to discuss again here: https://github.com/dojo/framework/issues/250). But my main questions are:

> This feature has been deprecated in favor of using [`Intl`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl) or the relevant [`Globalize.js`](https://github.com/globalizejs/globalize) methods directly.

I'd like to see whatever we plan on introducing with `Intl` land before these deprecations happen. Can the new `Intl` stuff we're replacing be polyfilled? If it can't be polyfilled then for me removing some of these api's should be off the table as it would be the only way to provide support across both legacy and modern.

To add to this if we're going to remove `Globalize` then I don't see why we're then going to recommend using it afterwards. If the answer to this is that because `Intl` is still not comprehensive then asking the user to mix both `Globalize` and `Intl` in places themselves is in my opinion a worse user experience than us keeping the existing api's and delegating to the appropriate where possible. If we really are removing `Globalize` i'm also not sure we should be providing native support for it in cli-build-app either, as that opens the door to a) version incompatibilities, and b) users requesting support for other external libraries in the build tool.

> `@dojo/framework/i18n` relies on [MessageFormat.js](https://messageformat.github.io/messageformat/) for [ICU message formatting](http://userguide.icu-project.org/formatparse/messages), which ships with its own up-to-date copy of the necessary CLDR data. As such, no additional CLDR data are required to use ICU message formatting in Dojo.

What is the bundle size of the CLDR data included with messageformat? Does this mean in the scenario where a user uses Globalize to say do unit formatting along side messageformat we're going to pay for similar CLDR content twice in bundle size?

agubler:
@mwistrand I'm going to go ahead and close this, at least for the time being given all the changes to i18n.

PR #662, opened 2020-02-01
Update to the creating widgets docs for defining a widgets primary key property

agubler:
**Type:** feature

**Description:**

Adds a section in the docs on `.key('id')` usage.

### DIFF
--- a/docs/en/creating-widgets/supplemental.md
+++ b/docs/en/creating-widgets/supplemental.md
@@ -345,9 +345,9 @@ Dojo uses a virtual node's key to uniquely identify a specific instance when re-
 
 ## Defining widget `key`s
 
-Traditionally the widget's `key` property is used by the Dojo rendering engine to uniquely identify and track widgets across renders. However updating the `key` property is also an effective way to guarantee that during the next render Dojo's rendering engine will recreate the widget instead of reusing the previous instance. In doing so this means all previous state will be reset widget and can be useful when working with widgets that manage logic based on a widget property.
+Traditionally the widget's `key` property is used by the Dojo rendering engine to uniquely identify and track widgets across renders. However, updating the `key` property is also an effective way to guarantee that during the next render Dojo's rendering engine will recreate the widget instead of reusing the previous instance. When recreating the widget all previous state will get reset. This behavior is useful when working with widgets that manage logic based on the value of a widget property.
 
-Dojo provides a mechanism for widget authors to associate a widget property to the widgets identity, by using the `.key()` chained method from the `create()` factory.
+Dojo provides a mechanism for widget authors to associate a widget property to the widget's identity by using the `.key()` chained method from the `create()` factory.
 
 ```tsx
 import { create } from '@dojo/framework/core/vdom';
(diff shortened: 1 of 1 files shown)


docs/en/creating-widgets/supplemental.md
```diff
@@ -343,6 +343,60 @@ Dojo uses a virtual node's key to uniquely identify a specific instance when re-
 
 > **Note:** Virtual node `keys` should be consistent across multiple render function invocations. Generating different keys for what should be the same output node within every render call [is considered an anti-pattern in Dojo application development](/learn/creating-widgets/best-practice-development#the-virtual-dom), and should be avoided.
 
+## Defining widget `key`s
+
+Traditionally the widget's `key` property is used by the Dojo rendering engine to uniquely identify and track widgets across renders. However updating the `key` property is also an effective way to guarantee that during the next render Dojo's rendering engine will recreate the widget instead of reusing the previous instance. In doing so this means all previous state will be reset widget and can be useful when working with widgets that manage logic based on a widget property.
```
dylans:
```suggestion
Traditionally the widget's `key` property is used by the Dojo rendering engine to uniquely identify and track widgets across renders. However, updating the `key` property is also an effective way to guarantee that during the next render Dojo's rendering engine will recreate the widget instead of reusing the previous instance. When recreating the widget all previous state will get reset. This behavior is useful when working with widgets that manage logic based on the value of a widget property.
```

docs/en/creating-widgets/supplemental.md
```diff
@@ -343,6 +343,60 @@ Dojo uses a virtual node's key to uniquely identify a specific instance when re-
...
+
+Dojo provides a mechanism for widget authors to associate a widget property to the widgets identity, by using the `.key()` chained method from the `create()` factory.
```
dylans:
```suggestion
Dojo provides a mechanism for widget authors to associate a widget property to the widget's identity by using the `.key()` chained method from the `create()` factory.
```

docs/en/creating-widgets/supplemental.md
```diff
@@ -343,6 +343,60 @@ Dojo uses a virtual node's key to uniquely identify a specific instance when re-
...
+interface MyWidgetProperties {
+	id: string;
+}
+
+const factory = create()
+	.properties<MyWidgetProperties>()
+	.key('id');
+```
+
+Using this factory Dojo will recreate the widget instance if the `id` property changes. This powerful feature provides widget authors assurance their widget will be recreated when the defined property changes, therefore not having to deal with often complicated logic to refresh data based on the property.
```
dylans:
```suggestion
Using this factory Dojo will recreate the widget instance if the `id` property changes. This powerful feature provides widget authors assurance their widget will get recreated when the defined property changes, therefore not having to deal with complicated logic to refresh data based on the property.
```

docs/en/creating-widgets/supplemental.md
```diff
@@ -343,6 +343,60 @@ Dojo uses a virtual node's key to uniquely identify a specific instance when re-
...
+
+	return (
+		<div>
+			<ul>{data.map((item) => <li>{item}</li>)}</ul>
+		</div>
+	);
+});
+```
+
+This example demonstrates fetching data based the `id` property, without using `.key('id)` the widget would need to manage scenarios where the `id` property changes. This would include logic to determine the property has actually changed, re-fetch the data and also show the loading message. Using `.key('id')` means that we know that when the `id` property changes the widget will be recreated and the state reset, so that the widget show the "Loading Data..." message and fetch data based on the updated `id`.
```
dylans:
```suggestion
This example demonstrates fetching data based on the `id` property. Without using `.key('id)`, the widget would need to manage scenarios where the `id` property changes. This would include logic to determine if the property has actually changed, re-fetch the relevant data and also show the loading message. Using `.key('id')` guarantees that when the `id` property changes the widget will get recreated and the state reset, and the widget shows the "Loading Data..." message and fetches data based on the updated `id`.
```

agubler:
@dylans Thanks, updated.

PR #672, opened 2020-02-06
Data middleware and resource

### LINKED ISSUE #660: Data middleware for function based widgets
This issue forms the proposal for a data middleware that widgets can use to become data aware and interact with an underlying shared data resource.

## Aims
- Consistent API
- Share data between widgets
- Share resources between widgets
- Silo resources so that they don't affect other widgets
- Reactive
- Pagination support
- Query support
- Data transform support
- Fully typed

## Data middleware POC achievements
The data middleware should provide the widget with an API to access and interact with the underlying data resource.

### passing resource to the middleware
The data middleware will add a `resource` property to the widget using it. This will allow a resource to be passed to the widget via a simple property. When initialising the data middleware, it should peel the resource off of the widget properties and bind it to the middleware api.

### passing a transformer to the middleware
The data middleware will need to provide a means to transform the data it is provided by it's resource in order the pass appropriate data to it's widget. This will be fully typed to ensure that the transformer that is passed via a transform property is suitable for the widget in question. In order to support these typings, we will need to provide a `createDataMiddleware` function similar to that which `icache` provides with `createIcacheMiddleware`.

```ts
// when creating a middleware for a DataWidget with this generic
const data = createDataMiddleware<{ foo: string; }>();

// types correctly
<DataWidget transform={(item) => ({ foo: item.firstName })} />

// should show compile error because it does not return `{foo: string}`.
<DataWidget transform={(item) => ({ bar: item.firstName })} />
```

### initialising the data middleware

The data middleware will provide a factory function that is used to initialise the middleware with options. During this initialisation the resource and a transform function where appropriate will be peeled off of the widget properties and bound to the data middleware.

The returned object can be destructured to gain access to the middleware API as described further below. This is how widgets will interact with the underlying data-resource.

When calling `data()` to initialise the middleware, options can be passed to effect the middleware that is created.

#### data middleware options

```ts
interface DataOptions {
  reset?: boolean;
  resource?: Resource | ResourceWrapper;
  key?: string;
}

interface Options {
  pageNumber: number;
  query: string;
  pageSize: number;
}
```

##### reset
When passing this option as `true` to the data middleware, the passed resource should be reset and created with it's own options. This means that if the widget is passed a `shared` resource, it will no longer be shared. This will be particularly useful in widgets such as the typeahead where we do not want changes to the query or the page to affect any other widgets elsewhere in the app that are using the same data resource.

##### resource
By default the middleware will take the resource from the widgets `resource` property, however, if you wish to override this or to work with more than one resource this will allow a user to create a second data api via the middleware to interact with it.

##### key
When working with the same resource in multiple places within the same widget we need to specify a `key` to differentiate them. Otherwise changing the page / query on one resource will have an impact on the other. 

### middleware api

```ts
interface DataMiddleware: {
    getOrRead(options: Options): DataResponse;
    loading(options: Options): boolean;
    failed(options: Options): boolean;
    setOptions(newOptions: Options): void;
    getOptions(): Options;
    readonly resource: ResourceWrapper;
    shared(): ResourceWrapper;
}
```

#### getOrRead
The get or read function will take a full options object including any query / pagination / pagesize etc information and will return the appropriate data. This should be a reactive function and cause the calling widget to invalidate when the data becomes available. This will use the underlying data-resource to acquire the data.

#### loading
We will need a way to discover if a particular data read is loading / in progress. This will enable us to display a loading indicator or progress bar where appropriate. This can be a sync api that returns a boolean to indicate the loading status of a `getOrRead`. As the `options` are used for all `getOrRead` commands, they can be used here too to ensure that the appropriate status is returned.

#### failed
We will need a way to discover if a getOrRead has failed. This can be achieved by passing the `options` object to the `failed` function. It can be a sync api the same as `loading` and can return a boolean response. An alternative to `loading` / `failed` might be to use a `status` api that returns a response of `'loading' | 'error' | 'complete'`. We should decide which of these is more appropriate.

#### setOptions
The setOptions function will take a set of options and use those to update the options stored in the middleware. This will enable the sharing of control over a resource between two or more widgets, for example a text input controlling the query on a menu.

#### getOptions
getOptions will return the current options object from the middleware. This will be used along with `getOrRead` to acquire the appropriate data. These could also be used for display purposes; ie. rendering out the query text.

#### get resource
The resource getter returns the underlying resource which can in turn be passed onto another data-aware widget. When they initialise their own data middleware with this resource, they will receive their own `options` layer, thus allowing them to interact with and query the same underlying data-resource via their own parameters.

#### shared
This function will return a shared resource which can then be passed to another widget. This shared resource will have the same options layer as the original and thus will share the sane query / pageNumber etc. This is particularly useful when passing a shared resource between a set of controls and another widget that renders the results.

---

A working POC of this proposal can be found [here](https://codesandbox.io/s/cold-wood-wmd0z)

tomdye:
**Type:** bug / feature

**Description:**

Adds a data middleware capable of making widgets data aware and able to deal with a resource.
- When using the middleware, `resource` will be added to the widget's properties. This can be used to pass a resource to be interacted with.
- Provides `createDataMiddleware` function which takes a generic to determine the type of data the widget requires, this adds a `transform` function to the widget properties which uses the generic to ensure the correct data type is supplied.

*Data middleware API interface:*

- getOrRead(options: ResourceOptions): any;
- getTotal(options: ResourceOptions): number | undefined;
- isLoading(options: ResourceOptions): boolean;
- isFailed(options: ResourceOptions): boolean;
- isLoading(options: ResourceOptions): boolean;
- isFailed(options: ResourceOptions): boolean;
- isLoading(options: ResourceOptions): boolean;
- isFailed(options: ResourceOptions): boolean;
- setOptions(newOptions: ResourceOptions): void;
- getOptions(): ResourceOptions;
- readonly resource: ResourceWrapper;
- shared(): ResourceWrapper;

`ResourceOptions` are made up of:
- pageNumber?: number
- pageSize?: number
- query?: string

When a widget using the middleware is destroyed, the middleware will disconnect from the resource.

*The Resource will need to provide the following API*

- getOrRead(options: ResourceOptions: any
- getTotal(options: ResourceOptions): number | undefined
- unsubscribe(invalidator: () => void): void
- subscribe(type: SubscriptionType, options: ResourceOptions, invalidator: () => void): void;
- isLoading(options: ResourceOptions): boolean;
- isFailed(options: ResourceOptions): boolean;

Both API 

Will work in future with resources.

Resolves #660 

Codesandbox of it in use with dummy resource: https://codesandbox.io/s/hardcore-swartz-kezpc

### DIFF
--- a/src/core/middleware/data.ts
+++ b/src/core/middleware/data.ts
@@ -160,6 +160,7 @@ export function createDataMiddleware<T = void>() {
 					nextResource.data.length !== currentResource.data.length
 				) {
 					resourceWithDataMap.delete(nextResource.data);
+					invalidator();
 				}
 			}
 		});
--- a/src/core/middleware/data.ts
+++ b/src/core/middleware/data.ts
@@ -3,20 +3,18 @@ import { create, invalidator } from '../vdom';
 type Invalidator = () => void;
 
 export interface ResourceOptions {
-	pageNumber: number;
-	query: string;
-	pageSize: number;
+	pageNumber?: number;
+	query?: string;
+	pageSize?: number;
 }
 
 export interface Resource {
-	getOrRead: (options: ResourceOptions) => any;
-	registerInvalidator: (invalidator: Invalidator) => void;
+	getOrRead: (options: ResourceOptions, invalidator: Invalidator) => any;
 }
 
 interface OptionsWrapper {
-	readonly options: ResourceOptions;
-	setOptions(newOptions: ResourceOptions): void;
-	registerInvalidator(invalidator: Invalidator): void;
+	getOptions(invalidator: Invalidator): ResourceOptions;
+	setOptions(newOptions: ResourceOptions, invalidator: Invalidator): void;
 }
 
 export interface ResourceWrapper {
@@ -50,11 +48,7 @@ function isDataTransformProperties<T>(
 }
 
 function createOptionsWrapper(): OptionsWrapper {
-	let options: ResourceOptions = {
-		pageNumber: 1,
-		pageSize: 10,
-		query: ''
-	};
+	let options: ResourceOptions = {};
 
 	const invalidators = new Set<Invalidator>();
 
@@ -65,17 +59,16 @@ function createOptionsWrapper(): OptionsWrapper {
 	}
 
 	return {
-		setOptions(newOptions: ResourceOptions) {
+		setOptions(newOptions: ResourceOptions, invalidator: Invalidator) {
+			invalidators.add(invalidator);
 			if (newOptions !== options) {
 				options = newOptions;
 				invalidate();
 			}
 		},
-		get options() {
-			return options;
-		},
-		registerInvalidator: (invalidator: Invalidator) => {
+		getOptions(invalidator: Invalidator) {
 			invalidators.add(invalidator);
+			return options;
 		}
 	};
 }
@@ -121,20 +114,18 @@ export function createDataMiddleware<T = void>() {
 			let cachedOptions = keyedCachedOptions.get(key);
 			let optionsWrapper: OptionsWrapper;
 
-			if (!cachedOptions) {
+			if (cachedOptions) {
+				optionsWrapper = cachedOptions;
+			} else {
 				const newOptionsWrapper = resourceWrapper.createOptionsWrapper();
 				keyedCachedOptions.set(key, newOptionsWrapper);
 				optionsWrapperMap.set(resource, keyedCachedOptions);
 				optionsWrapper = newOptionsWrapper;
-			} else {
-				optionsWrapper = cachedOptions;
 			}
 
-			// Return the data API
 			return {
 				getOrRead(options: ResourceOptions) {
-					resource.registerInvalidator(invalidator);
-					const data = resource.getOrRead(options);
+					const data = resource.getOrRead(options, invalidator);
 					const props = properties();
 
 					if (data && data.length && isDataTransformProperties(props)) {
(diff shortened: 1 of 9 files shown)


tests/core/unit/middleware/data.tsx
```diff
@@ -0,0 +1,254 @@
...
+	registerInvalidator: sb.stub()
+};
+
+jsdomDescribe('data middleware', () => {
+	beforeEach(() => {
+		resolvers.stub();
+	});
+	afterEach(() => {
+		resolvers.restore();
+		sb.resetHistory();
```
matt-gadd: do we use this abbreviation elsewhere? `sb`?

src/core/middleware/data.ts
```diff
@@ -0,0 +1,168 @@
...
+
+export interface ResourceOptions {
+	pageNumber: number;
+	query: string;
+	pageSize: number;
+}
+
+export interface Resource {
+	getOrRead: (options: ResourceOptions) => any;
+	registerInvalidator: (invalidator: Invalidator) => void;
```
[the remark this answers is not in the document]
tomdye: good call, thanks

tests/core/unit/middleware/data.tsx
tomdye: believe so

tomdye: Screenshot 2020-02-07 at 17 00 48
everywhere

agubler:
For APIs that do not provide a `total` the current implementation can be problematic, which means they will become less useful for using externally controlled/managed APIs. For example without a real total, returning a partial last page or an empty array (for APIs where there are no results) causes an infinite loop calling the user resource. It would be good is we could make the total option in some way to support this use case.

Additionally it would be beneficial if the data api could be used without providing pagination details, for example if a user wants to return a single resource.

Here is an implementation of the HNPWA using data/resources that demonstrates these two scenarios https://github.com/agubler/dojo-resource-hnpwa

The resources are defined in `App.tsx` https://github.com/agubler/dojo-resource-hnpwa/blob/master/src/App.tsx

PR #674, opened 2020-02-11
An empty computed locale causes message values to be undefined

agubler:
**Type:** bug

**Description:**

Sets a default for the `computedLocale` as the existing `''` causes messages to return as undefined. This did not surface in testing as it the computed locale was always being set via `setLocale` however when using this downstream with dojo/widgets the tests were not able to return the default message bundles.

### DIFF
--- a/src/i18n/i18n.ts
+++ b/src/i18n/i18n.ts
@@ -72,6 +72,8 @@ const DOJO_PATH = 'dojo/{bundle}/lookup';
 
 let supportedLocales: string[] = [];
 let defaultLocale = '';
+// Set to `unknown` to support using default message bundles
+// without an application locale configured
 let computedLocale = 'unknown';
 let currentLocale = '';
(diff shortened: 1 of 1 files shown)


src/i18n/i18n.ts
```diff
@@ -72,7 +72,7 @@ const DOJO_PATH = 'dojo/{bundle}/lookup';
 
 let supportedLocales: string[] = [];
 let defaultLocale = '';
-let computedLocale = '';
+let computedLocale = 'unknown';
```
agubler: Happy to go with `default` / `unset` / `unspecified` or even use the browsers default i.e. `global.navigator.language || global.navigator.userLanguage`

This is only going be used when there is no `locale` set in the `.dojorc` and therefore be used for loading the default messages from nls bundles.

mwistrand: I'm fine with `unknown`, but a quick comment explaining its purpose would be nice.

PR #680, opened 2020-02-20
Unmount Application

### LINKED ISSUE #676: Fully unmounting entire application
Hi there 👋 

I'm the maintainer of [single-spa](https://single-spa.js.org), the most popular open source microfrontends framework.

We recently created https://github.com/single-spa/single-spa-dojo, which is the adapter layer needed for dojo applications to participate in a single-spa microfrontends web app.

To do so, we need to be able to fully unmount a Dojo application, including calling all `onDetach` / `destroy` lifecycles. I have been unable to find a way to do this in the dojo documentation and source code. As a workaround we are doing `renderer(<div style="display: none"></div>)` on the same div where dojo is mounted. The problem is that this does not actually call the onDetach lifecyle. You can see more detail at https://github.com/single-spa/single-spa-dojo/issues/5

The closest thing I've been able to find in the source code for this is the `DeleteDomApplication` interface inside of vdom.ts: https://github.com/dojo/framework/blob/824a7b3aed3ce9422f2991ae5cbeb156ccedbcba/src/core/vdom.ts#L1766-L1773

Is full unmounting possible?

agubler:
**Type:** feature

**Description:**

Adds `unmount` to the `renderer` to fully detached and unmount an entire application.

Resolves #676

### DIFF
--- a/src/core/vdom.ts
+++ b/src/core/vdom.ts
@@ -126,6 +126,7 @@ export interface MountOptions {
 export interface Renderer {
 	invalidate(): void;
 	mount(mountOptions?: Partial<MountOptions>): void;
+	unmount(): void;
 }
 
 interface ProcessItem {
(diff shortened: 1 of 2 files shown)


joeldenning:
Amazing - thanks. Looking forward to this being released. One single-spa user is on dojo@6 so I think they would appreciate this being backported to dojo 6 if possible too.

I appreciate the prompt response and your work on this.

Issue #683, opened 2020-03-10
Theme Variants


tomdye:
### Problem

Currently we use css-variables to share colours / sizes etc across the theme and when we build these we both inline and bullet proof these variables into the generated CSS. css-variables are currently defined at the `:root` level and imported into each of the css files that uses them. This means that we end up with multiple copies of the css-variables in the css bundle and that we are unable to override these variables easily to change the appearance of a theme as we cannot unload the previous set of variables or easily control the specificity of them.

With our current configuration, to build theme variants we would need to build multiple variants of our themes with different variables, this would create large css files which we would like to avoid. 

We would like to support setting both the `theme` and the `variant` at both a global and a widget level. This would allow us to follow the same pattern we currently use for either injecting or prop-passing themes in order the pass variants.

We need to ensure discoverability of theme variants so that a user can import and use them easily.

### Proposal

In order to enable theme variants we should separate out the css and the css-variables from our themes so that we can support theme variants by simply changing the set of variables we supply. This would mean that we would no longer support the generic `variables.css` file we currently do and would not include these variables in our theme builds.

In order to support theme variants via css-variables we could convert our variables files into css-modules that contain variables within a class. This class can then be applied to a domnode in order to cascade those variables to any domnodes beneath which are using them.

We can support this by making changes to the theme middleware (and mixin) to return a variant class which widgets can then apply to their root to support a variant. Furthermore we should change our theme index file signature to return a variants object which can be used to aid discovery and setting of a default variant. 

#### framework

We will need to make changes to the existing theme middleware / mixin to support setting a theme variant and retrieving a class that applies the current variant. We will need to deal with variant defaults also.

#### IE11

In order to support IE11 we will need to use a polyfill for run time css-variables. A suitable poly fill can be found [here](https://www.npmjs.com/package/ie11-custom-properties).

#### cli-build-theme

We will need to ensure that cli-build-theme no longer looks for and inlines a `variables.css` file. This has always been a _special case_ that cli-build-theme had to cater for in order to support inlining of css variables. We will need to remove our existing bulletproofing fallbacks in order to ensure ie11 behaves as expected and to ensure that variant variables always override with expected specificity.

#### existing themes

Existing themes will need to have their `variables.css` converted to a variant `.m.css` file and export a default variant from their index file. We will need to remove all imports of `variables.css` from our themes. Some folder structure changes may be required

#### widgets

[...]

PR #684, opened 2020-03-11
Add Theme variants

tomdye:
**Type:** bug / feature

**Description:**

Adds ability to pass both a theme and a variant to widgets via theme middleware / mixin

Related to: https://github.com/dojo/framework/issues/683

### DIFF
--- a/src/core/interfaces.d.ts
+++ b/src/core/interfaces.d.ts
@@ -124,6 +124,23 @@ export interface Theme {
 	[key: string]: object;
 }
 
+export interface Variant {
+	root: string;
+}
+
+export interface ThemeVariant {
+	theme: Theme;
+	variant: Variant;
+}
+
+export interface ThemeVariantConfig {
+	theme: Theme;
+	variants: {
+		default: Variant;
+		[key: string]: Variant;
+	};
+}
+
 export interface Classes {
 	[widgetKey: string]: {
 		[classKey: string]: SupportedClassName[];
--- a/src/core/interfaces.d.ts
+++ b/src/core/interfaces.d.ts
@@ -129,8 +129,8 @@ export interface Variant {
 }
 
 export interface ThemeVariant {
-	theme: Theme;
-	variant: Variant;
+	theme: Theme | ThemeVariantConfig;
+	variant: Variant | string;
 }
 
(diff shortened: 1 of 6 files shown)


src/core/interfaces.d.ts
```diff
@@ -124,6 +124,23 @@ export interface Theme {
 	[key: string]: object;
 }
 
+export interface Variant {
+	root: string;
+}
+
+export interface ThemeVariant {
```
agubler: Would this be better as `ThemeWithVariant` and then the interface below `ThemeWithVariants`?

src/core/interfaces.d.ts
```diff
@@ -124,6 +124,23 @@ export interface Theme {
...
+	theme: Theme | ThemeVariantConfig;
```
agubler: Do you think `theme` should now be called `css`? Would make it less weird when accessing them, so would be `theme.css` over `theme.theme`

src/core/middleware/theme.ts
```diff
@@ -10,14 +10,26 @@ import Registry from '../Registry';
...
 
 export const THEME_KEY = ' _key';
 
 export const INJECTED_THEME_KEY = '__theme_injector';
 
+function isThemeVariant(theme: Theme | ThemeVariant): theme is ThemeVariant {
+	return theme.hasOwnProperty('variant');
+}
+
+function isThemeVariantConfig(theme: Theme | ThemeVariantConfig): theme is ThemeVariantConfig {
```
agubler: perhaps `isThemeWithVariants`? or `hasVariants`

src/core/middleware/theme.ts
```diff
@@ -100,14 +137,32 @@ export const theme = factory(
 				icache.set(css, theme, false);
 				return theme;
 			},
-			set(css: Theme): void {
-				const currentTheme = injector.get<Injector<Theme | undefined>>(INJECTED_THEME_KEY);
-				if (currentTheme) {
-					currentTheme.set(css);
+			variant() {
+				let theme;
+				const { theme: themeProp } = properties();
```
agubler: The theme property should be either the explicitly passed theme or the injected theme already, so can just use the property?

PR #705, opened 2020-03-27
Update i18n docs to reflect v7 changes

### LINKED ISSUE #698: Update i18n Reference Docs
**Enhancement** 

The changes to the `i18n` support in Dojo 7 means that the reference guide needs updating to reflect the latest support.

### Specifying a root locale within an application 

Update this section to talk about the computed default locale being used as the default locale for the application. This depends on the locale and supportLocales specified in the applications `.dojorc` and the users browser locale.

### Lazy vs Static Loading

A function that returns an object is no longer support for locale bundles, it is either a promise factory or an object.

### Advanced formatting: CLDR

The new functionality automatically includes the CLDR data required based on the locales specified in the `.dojorc` and the usage of formatters such as date, number etc from i18n.

The CLDR is bundled separately and only loaded when required by the application, the user should never need to decide what CLDR is required anymore.

mwistrand:
**Type:** feature

**Description:**

Resolves #698

Unfortunately, as I cannot read Chinese, the corresponding `zh-CN` docs are left untouched.

- Replace references of the old `i18n` function with `localizeBundle()`.
- Remove references to CLDR data as these data are now injected at build time.
- Remove references to `setLocaleMessages` and `getCachedMessages`.
- Replace message formatter examples with `format` examples.
- Update documentation on reading the current locale.


src/i18n/i18n.ts
```diff
@@ -100,16 +100,16 @@ export function setDefaultLocale(locale: string) {
  * Returns the users locale computed by using the system locale
  * of the environment and the default locale.
  *
- * The users system local if supported by the application (i.e resolves
+ * The users system locale if supported by the application (i.e resolves
```
maier49:
```suggestion
 * The user's system locale if supported by the application (i.e resolves
```

xiaohulu:
@mwistrand I will update zh-CN docs.

agubler:
@mwistrand Do you think we need to keep a section on the build tooling? Covering the `locale` and `supportedLocales` options, as this is what is used to automatically inject the required CLDR data?

mwistrand:
@agubler There is already a "Configuring supported application locales" section beneath the top-level "Internationalizing a Dojo application" page that describes the `locale` and `supportedLocales` values in the `.dojorc.`

agubler:
Should we mention on that page that without the locale then i18n will only use root bundles and do non icu message formatting? Or do we think that's enough?

mwistrand:
That's not a bad idea. I'll update to include that warning.

PR #707, opened 2020-03-27
Allow generic-ish properties for functional widgets

### LINKED ISSUE #679: Support for generic parameters on functional widget properties
**Enhancement**

Class based widgets support generic type arguments for the widget's properties type. For instance:

```tsx
export interface ListProperties<T> {
    data: T[];
}

export class List<T> extends WidgetBase {
}
```

We should investigate adding this support for function-based widgets as well. The `icache` middleware does something similar with `createICacheMiddleware`.

maier49:
**Type:** feature

**Description:**
An attempt at resolving the issue of generic types for functional widgets. It's not ideal because you have to specify the entire interface again but it does allow you to narrow a generic down. I can't seem to think of a way to get more ideal typing to work without higher order types or creating the widget factory at at render which is obviously a nonstarter.

Resolves #679

### DIFF
--- a/src/core/interfaces.d.ts
+++ b/src/core/interfaces.d.ts
@@ -451,9 +451,12 @@ export interface MiddlewareResultFactory<Props, Children, Middleware, ReturnValu
 }
 
 export interface DefaultChildrenWNodeFactory<W extends WNodeFactoryTypes> {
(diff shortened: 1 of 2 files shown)


maier49:
Closing as the preferred approach would be to create a wrapping generic function to instantiate the a widget with the generic type baked in.

PR #710, opened 2020-03-31
Dojo Test Renderer

### LINKED ISSUE #529: Better Harness and Assertion Template support for render props
**Enhancement** 
Assertion Templates have made creating/modifying assertions a lot easier for testing, but it's still really difficult to call and assert against render props, which mostly involves abusing the `trigger` func and asserting raw `DNode`s rather than using an Assertion Template. It would be nice to improve the ergonomics of this so a user can write these the same as any other Assertion Template.

This might be more achievable with: https://github.com/dojo/framework/issues/528

### LINKED ISSUE #535: Rewrite of the test harness
**Enhancement** 
The testing harness is one of the oldest parts of the codebase, and since then a number of the new patterns have emerged such as render props that it doesn't deal very well with. In Dojo 7 we'd like to revisit the harness and see how we can improve the ergonomics as well as offering more options in terms of rendering.

Some goals:
* Expose less of the underlying DNodes to the end user, in favour of providing better mechanisms to trigger and expect assertions with.
* Support the existing virtual rendering (no vdom at all) as well as shallow, and full rendering to actual dom.
* First class support for render props
* Better debugging output when assertions fail

agubler:
### Type: feature

### Description:

Introduces a new testing module, `@dojo/framework/testing/renderer`,  intended to retire the existing `@dojo/framework/testing/harness`, the `harness` has been deprecated but not removed. 

The existing `harness` is has been moved to the `framework/testing/harness` folder.

#### Features:

* Refined the existing `harness` API to promote testing best practices (removed `expectPartial` and `getRender`).
* Introduces "wrapped test node" in order to interact with the assertion template and test renderer APIs instead of string selectors.
* Built in Type-safe support for triggering widget properties.
* Built in type-safe support for testing functional children.
* Makes assertion templates a first class citizen, the only way to test widgets.
* Type-safe assertion template API.
* Inline custom comparators using the `compare` function in combination with a wrapped testing node
* Output diffs in `tsx` format over `v()` and `w()`

```tsx
import {,tsx } from '@dojo/framework/core/vdom';
import renderer, { wrap, compare } from '@dojo/framework/testing/renderer';
import assertionTemplate from '@dojo/framework/testing/assertionTemplate';

import MyWidget from './MyWidget';
import Button from './Button';
import Card from './Card';

// create a wrapped node for anything that needs
// to be used with the renderer or assertion template api
const WrappedButton = wrap(Button);
const WrappedRoot = wrap('div');
const WrappedCard = wrap(Card);

// create the base template with the wrapped nodes
// in place of the real node/widget
const baseTemplate = assertionTemplate(() => (
    // use `compare` with any property on a wrapped node
    // to pass a custom comparator
    <WrappedRoot id={compare((actual) => typeof actual === 'string'}>
        <WrappedButton onClick={() => {}} />Show</WrappedButton>
    </WrappedRoot>
));

// initialise the test renderer
const r = renderer(() => <MyWidget />);

// always expect the default render state
r.expect(baseTemplate);

// register calling a property using a wrapped node
r.property(WrappedButton, 'onClick');
// describe how to resolve functional children
r.child(WrappedCard, [{ header: [], content: [] }]);
// create a new template using the wrapped nodes
// to target the changes required
const cardTemplate = baseTemplate
    .setProperty(WrappedRoot, 'classes', [css.open])
    .insertAfter(WrappedButton, () => (
        <Card>{{
            header: () => <h1>Header</h1>,
            content: () => <div>Content</div>
        }}
        </Card>
    ));
// expect the against the updated template
// the properties will be called before a the
// widget re-renders
r.expect(cardTemplate);

r.property(WrappedButton, 'onClick');
r.expect(baseTemplate);
```

**Note:** Testing render props is not possible with the new test renderer and recommended to migrate these widgets to use functional children for full test renderer support.

Resolves #535 
Resolves #529


docs/en/testing/supplemental.md
```diff
@@ -496,15 +494,18 @@ import { assert } from 'chai';
...
+				<WrappedButton key="button" onClick={() => {}}>
 					Fetch
-				</Button>
+				</WrappedButton>
 			</div>
 		));
-		h.trigger('@button', 'onClick');
+		const r = renderer(() => <Action fetchItems={fetchItems} />);
+		r.expect(template);
+		r.property(WrappedButton, 'onClick');
```
devpaul, line 507: It might be worth mentioning in the documentation below how the `onClick` property is found in the renderer using the wrapped button. It's not immediately clear how the `property` method is using the `WrappedButton` to find the real `Button` on the rendered widget when the rendered widget isn't constructed using the wrapped widgets.

devpaul:
What happens if a wrapped widget is used more than once?

```typescript
const WrappedButton = wrap(Button);
		h.expect(() => (
		const template = assertionTemplate(() => (
			<div classes={[css.root]}>
				<WrappedButton key="button1" onClick={() => {}}>
					Fetch
				</WrappedButton>
				<WrappedButton key="button2" onClick={() => {}}>
					Fetch
				</WrappedButton>
			</div>
		));
		const r = renderer(() => <MyWidget buttonActions={[ jest.fn(), jest.fn() ]} />);
		r.expect(template);
		r.property(WrappedButton, 'onClick');
```

Let's say that `MyWidget` has two buttons exactly like the assertion template and `buttonActions` are assigned respectively to each `Button`'s `onclick`. Are both mock functions called? Is an error produced? What happens?

docs/en/testing/introduction.md
```diff
@@ -41,7 +41,7 @@ dojo test --functional --config local
 
 ## Writing unit tests
 
--   Using Dojo's [`harness` API](/learn/testing/dojo-test-harness#harness-api) for unit testing widgets.
+-   Using Dojo's test [`renderer` API](/learn/testing/test-renderer) for unit testing widgets.
```
dylans, line 44:
```suggestion
-   Using Dojo's [test `renderer` API](/learn/testing/test-renderer) for unit testing widgets.
```

docs/en/testing/introduction.md
```diff
@@ -149,53 +148,66 @@ export default Profile;
...
+		const r = renderer(() => <Profile />);
 		// Test against my base assertion
-		h.expect(profileAssertion);
+		r.expect(profileAssertion);
 	});
 });
 ```
 
-A value can be provided to any virtual DOM node under test using `assertion-key` properties defined in the assertion template. Note: when `v()` and `w()` from `@dojo/framework/core/vdom` are used, the `~key` property serves the same purpose.
+To work with assertion templates, wrapped nodes can be created using `@dojo/framework/testing/renderer#wrap` in order to use the assertion template API. Note: when using wrapped `VNode`s with `v()`, the `.tag` property needs to be used, for example `v(WrappedDiv.tag, {} [])`.
```
dylans, line 172:
```suggestion
To work with assertion templates, wrapped nodes can get created using `@dojo/framework/testing/renderer#wrap` in order to use the assertion template API. Note: when using wrapped `VNode`s with `v()`, the `.tag` property needs to get used, for example `v(WrappedDiv.tag, {} [])`.
```

docs/en/testing/introduction.md
```diff
@@ -149,53 +148,66 @@ export default Profile;
...
+
+import Profile from '../../../src/widgets/Profile';
+import * as css from '../../../src/widgets/Profile.m.css';
+
+// Create a wrapped test node
+const WrappedHeader = wrap('h1');
+
+// Create an assertion
+const profileAssertion = assertionTemplate(() => (
+	// use the wrapped node in place of the normal node
```
dylans, line 190:
```suggestion
	// Use the wrapped node in place of the normal node
```

agubler:
> What happens if a wrapped widget is used more than once?
> 
> ```ts
> const WrappedButton = wrap(Button);
> const template = assertionTemplate(() => (
> 	<div classes={[css.root]}>
> 		<WrappedButton key="button1" onClick={() => {}}>
> 			Fetch
> 		</WrappedButton>
> 		<WrappedButton key="button2" onClick={() => {}}>
> 			Fetch
> 		</WrappedButton>
> 	</div>
> ));
> const r = renderer(() => <MyWidget buttonActions={[ jest.fn(), jest.fn() ]} />);
> r.expect(template);
> r.property(WrappedButton, 'onClick');
> ```
> 
> Let's say that `MyWidget` has two buttons exactly like the assertion template and `buttonActions` are assigned respectively to each `Button`'s `onclick`. Are both mock functions called? Is an error produced? What happens?

@devpaul at the moment it would actually call the corresponding `onClick` of both the button widgets in the widget under tests rendered output. This actually feels incorrect to me and it should only call one, basically you shouldn't use the same wrapped node twice in the assertion structure. If we change to follow that rule then it would only call the first wrapped node it finds and no more. This would then lead to an assertion failure in the test as it wouldn't have called the second `onClick`

docs/en/testing/supplemental.md
agubler: That seems reasonable, I'll add something into the supplemental docs.

agubler: @devpaul I have added:

> The test renderer uses the location of a wrapped test node in the expected tree to attempt to perform the requested action (either `r.property()` or `r.child()`) on the actual output of the widget under test. If the wrapped test node does not match the corresponding node in the actual output tree then no action will be performed and the assertion will report a failure.

To the `Wrapped Test Nodes` section.

devpaul:
> This actually feels incorrect to me and it should only call one, basically you shouldn't use the same wrapped node twice in the assertion structure

Sounds like it would be reasonable to throw an exception and fail the test at this point. It would give us a chance to provide a good error message that educates the user on proper usage. If we proceed there's a chance that it might work for their test (i.e. maybe they only check the first value).

agubler:
> Sounds like it would be reasonable to throw an exception and fail the test at this point. It would give us a chance to provide a good error message that educates the user on proper usage. If we proceed there's a chance that it might work for their test (i.e. maybe they only check the first value).

@devpaul the test renderer will now throw an error if the same wrapped test node is detected during an assertion

docs/en/testing/supplemental.md
devpaul: Thanks for adding this. It'll be a good reference!

docs/en/testing/supplemental.md
```diff
@@ -1,378 +1,454 @@
...
+const WrappedMyWidget = wrap(MyWidget);
+
+// Create a wrapped node for a vnode
+const WrappedDiv = wrap('div');
 ```
 
--   `renderFunction`: A function that returns a `WNode` for the widget under test
--   [`customComparators`](/learn/testing/dojo-test-harness#custom-comparators): Array of custom comparator descriptors. Each provides a comparator function to be used during the comparison for `properties` located using a `selector` and `property` name
--   `options`: Expanded options for the harness which includes `customComparators` and an array of middleware/mocks tuples.
+The test renderer uses the location of a wrapped test node in the expected tree to attempt to perform the requested action (either `r.property()` or `r.child()`) on the actual output of the widget under test. If the wrapped test node does not match the corresponding node in the actual output tree then no action will be performed and the assertion will report a failure.
```
maier49, line 50: It seems like this statement is made out of context here

docs/en/testing/supplemental.md
```diff
@@ -1,378 +1,454 @@
-# Dojo test harness
+# Test Renderer
 
-`harness()` is the primary API when working with `@dojo/framework/testing`, essentially setting up each test and providing a context to perform virtual DOM assertions and interactions. The harness is designed to mirror the core behavior for widgets when updating `properties` or `children` and widget invalidation, with no special or custom logic required.
+Dojo provides a simple and type safe test renderer for shallowly asserting the expected output and behavior from a widget. The test renderer's API has been designed to encourage unit testing best practices from the outset to ensure high confidence in you Dojo application.
 
-## Harness API
+Working with [assertion templates](/missing/link) and the test renderer is done using [wrapped test nodes](/missing/link) that are defined in the assertion templates structure, ensuring type safety throughout the testing life-cycle.
```
maier49, line 5: Are these links coming in later PRs?

docs/en/testing/supplemental.md
```diff
@@ -1,378 +1,454 @@
...
+```tsx
+.setChildren(
+  wrapped: Wrapped,
+  children: () => RenderResult,
+  type: 'prepend' | 'replace' | 'append' = 'replace'
+);`
+```
 
-The descriptors have a [`selector`](/learn/testing/dojo-test-harness#selectors) to locate the virtual nodes to check, a property name for the custom compare and a comparator function that receives the actual value and returns a boolean result for the assertion.
+Convience functions exists for all 3 types, [`prepend()`](/missing/link), [`append()`](/missing/link) and [`replaceChildren()`](/missing/link).
```
maier49, line 165:
```suggestion
Convenience functions exists for all 3 types, [`prepend()`](/missing/link), [`append()`](/missing/link) and [`replaceChildren()`](/missing/link).
```

docs/en/testing/supplemental.md
```diff
@@ -1,378 +1,454 @@
...
--   `invalidator`
--   `setProperty`
--   `destroy`
+### Assertion Template API
 
-Additionally, there are a number of mock middleware available to support widgets that use the corresponding provided Dojo middleware. See the [mocking](/learn/testing/mocking#provided-middleware-mocks) section for more information on provided mock middleware.
+#### `assertionTemplate.setChildren()`
 
-## Custom comparators
+Returns a new assertion template with the new children are either pre-pended, appended or replaced depending on the `type` passed.
```
maier49, line 155:
```suggestion
Returns a new assertion template with the new children either pre-pended, appended or replaced depending on the `type` passed.
```

docs/en/testing/supplemental.md
```diff
@@ -1,378 +1,454 @@
...
 
-```ts
-const compareId = {
-	selector: '*', // all nodes
-	property: 'id',
-	comparator: (value: any) => typeof value === 'string' // checks the property value is a string
-};
+#### `assertionTemplate.append()`
+
+Returns a new assertion template with the new children appended to the nodes existing children.
```
maier49, line 169:
```suggestion
Returns a new assertion template with the new children appended to the node's existing children.
```

docs/en/testing/supplemental.md
```diff
@@ -1,378 +1,454 @@
...
 
-const h = harness(() => w(MyWidget, {}), [compareId]);
+```tsx
+.append(wrapped: Wrapped, children: () => RenderResult);
 ```
 
-For all assertions, using the returned `harness` API will now only test identified `id` properties using the `comparator` instead of the standard equality.
+#### `assertionTemplate.prepend()`
+
+Returns a new assertion template with the new children pre-pended to the nodes existing children.
```
maier49, line 177:
```suggestion
Returns a new assertion template with the new children pre-pended to the node's existing children.
```

docs/en/testing/supplemental.md
```diff
@@ -1,378 +1,454 @@
...
-## Selectors
+```tsx
+.append(wrapped: Wrapped, children: () => RenderResult);
+```
 
-The `harness` APIs commonly support a concept of CSS style selectors to target nodes within the virtual DOM for assertions and operations. Review the [full list of supported selectors](https://github.com/fb55/css-select#supported-selectors) for more information.
+#### `assertionTemplate.replaceChildren()`
 
-In addition to the standard API:
+Returns a new assertion template with the new children replacing the nodes existing children.
```
maier49, line 185:
```suggestion
Returns a new assertion template with the new children replacing the node's existing children.
```

docs/en/testing/supplemental.md
```diff
@@ -1,378 +1,454 @@
...
--   The `classes` property is used instead of `class` when using the standard shorthand `.` for targeting classes
+```tsx
+.append(wrapped: Wrapped, children: () => RenderResult);
+```
 
-## `harness.expect`
+#### `assertionTemplate.insertSiblings()`
 
-The most common requirement for testing is to assert the structural output from a widget's `render` function. `expect` accepts a render function that returns the expected render output from the widget under test.
+Returns a new assertion template with the passed children either inserted either `before` or `after` depending on the `type` passed.
```
maier49, line 193:
```suggestion
Returns a new assertion template with the passed children inserted either `before` or `after` depending on the `type` passed.
```

docs/en/testing/supplemental.md
```diff
@@ -1,378 +1,454 @@
...
 
-```ts
-h.expect(() =>
-	<div key="foo">
-		<Widget key="child-widget" />
-		text node
-		<span classes={[class]} />
-	</div>
-);
+Returns a new assertion template with the passed children inserted before the existing nodes children.
```
maier49, line 205:
```suggestion
Returns a new assertion template with the passed children inserted before the existing node's children.
```

docs/en/testing/supplemental.md
```diff
@@ -1,378 +1,454 @@
...
+```tsx
+.insertBefore(wrapped: Wrapped, children: () => RenderResult);
 ```
 
-Optionally `expect` can accept a second parameter of a function that returns a render result to assert against.
+#### `assertionTemplate.insertAfter()`
 
-```ts
-h.expect(() => <div key="foo" />, () => <div key="foo" />);
+Returns a new assertion template with the passed children inserted after the existing nodes children.
```
maier49, line 213:
```suggestion
Returns a new assertion template with the passed children inserted after the existing node's children.
```

docs/en/testing/supplemental.md
```diff
@@ -1,378 +1,454 @@
...
+
+```tsx
+.setProperties<T>(
+  wrapped: Wrapped<T>,
+  value: T['properties'] | PropertiesComparatorFunction<T['properties']>
+): AssertionTemplateResult;
 ```
 
-A `functionalSelector` can be used return a function that is nested in a widget's properties. The function will be triggered, in the same way that using a plain string `functionSelector`.
+A function can be set in place of the properties object to return the expected properties based of the actual properties.
```
maier49, line 257:
```suggestion
A function can be set in place of the properties object to return the expected properties based off the actual properties.
```

docs/en/testing/supplemental.md
```diff
@@ -1,378 +1,454 @@
...
 
-### Trigger example
+## Triggering Properties
 
-Given the following VDOM structure:
+In addition to asserting the output from a widget, widget behavior can be tested by using the `renderer.property()` function. The `property()` function takes a [wrapped test node]() and the key of a property to call before the next call to `expect()`.
```
maier49, line 261: It can take additional arguments to pass to the function still right? Maybe worth mentioning

docs/en/testing/supplemental.md
agubler: @maier49 It's trying to explain how the wrapped nodes are used, is there a more appropriate place outside of the test node section?

docs/en/testing/supplemental.md
agubler: Updated

docs/en/testing/supplemental.md
agubler: Updated

PR #714, opened 2020-04-02
Custom element slots

### LINKED ISSUE #708: Custom Element support for child functions and named child functions
**Enhancement** 
In Dojo 7 we now have more types of children for widgets, the existing scenario being:

```tsx
<Foo>
    <Bar />
</Foo>
```

With the new children functionality we now have the following additional ways to accept children:

Render func:
```tsx
<Foo>
    { () => <Bar /> }
</Foo>
```

Named renders:
```tsx
<Foo>
    {{ 
      bar: <Bar />, 
      qux: <Qux /> 
    }}
</Foo>
```

Named render funcs:
```tsx
<Foo>
    {{ 
      bar: () => <Bar />, 
      qux: () => <Qux /> 
    }}
</Foo>
```
Named render funcs that inject values:
```tsx
<Foo>
    {{ 
      bar: (active) => <Bar active={ active } />, 
      qux: (active) => <Qux active={ active } /> 
    }}
</Foo>
```

None of the above are supported with Custom Elements.

Some thoughts on how we could at least add basic support for the first 3:

* To support basic render functions, we could add a function on top of our `VNode` and `WNode` objects, that return them themselves. This would mean we would not need to touch the Custom Element code for this support, as basically any Node can double as a factory safely. This probably also gives the benefit of every child being able to support both `RenderResult` and `() => RenderResult`, without any additional work.

* For the named children. We add an additional property to our Custom Element wrapper that allows the user to specify the child name as a property. This would be similar to `slot`'s for Custom Elements and we might even want to use that name.

An example of the above, for the following widget use in dojo:
```tsx
<Foo>
    {{ 
      bar: <Bar />, 
      qux: <Qux /> 
    }}
</Foo>
```

for custom elements it would be:
```html
<dojo-foo>
  <dojo-bar slot="bar" />
  <dojo-qux slot="qux" />
</dojo-foo>
```

The final scenario ie passing things into a render func, will need further thought as it is not suited to an event handler, and it means that the child nodes will essentially no longer be able to be used declaratively.

rorticus:
**Type:** feature

**Description:**

Adding custom element support for (some) child functions.  

The following scenarios are supported now,

### Dojo
```tsx
<Foo>
    { () => <Bar /> }
</Foo>
```

### Custom Elements

```html
<element-foo>
    <element-bar></element-bar>
</element-foo>
```

### Dojo

```tsx
<Foo>
    {{ 
      bar: <Bar />, 
      qux: [<Qux />, <Qux />]
    }}
</Foo>
```

### Custom Elements

```html
<element-foo>
    <element-bar slot="bar"></element-bar>
    <element-qux slot="qux"></element-qux>
    <element-qux slot="qux"></element-qux>
</element-foo>
```

### Dojo

```tsx
<Foo>
    {{ 
      bar: () => <Bar />, 
      qux: () => <Qux /> 
    }}
</Foo>
```

### Custom Elements

```html
<element-foo>
    <element-bar slot="bar"></element-bar>
    <element-qux slot="qux"></element-qux>
</element-foo>
```

### Dojo

```tsx
<Foo>
{{
    bar: (active) => <Bar active={active} />
}}
</Foo>
```

### Custom Elements

```html
<element-foo>
    <element-bar id="bar" slot="bar"></element-bar>
</element-foo>

<script>
bar.addEventListener('render', function ({ detail: [active]}) {
    this.active = active;
});
</script>
```

Resolves #708

### DIFF
--- a/src/core/registerCustomElement.ts
+++ b/src/core/registerCustomElement.ts
@@ -1,8 +1,9 @@
 import Registry from './Registry';
-import { renderer, w, dom, isTextNode, create as vdomCreate, diffProperty, invalidator } from './vdom';
+import { create as vdomCreate, diffProperty, dom, invalidator, isTextNode, renderer, w } from './vdom';
 import { from } from '../shim/array';
 import global from '../shim/global';
 import Injector from './Injector';
+import { VNode, WNode } from './interfaces';
 
 const RESERVED_PROPS = ['focus'];
 
@@ -246,6 +247,16 @@ export function create(descriptor: any, WidgetConstructor: any): any {
 		}
 
 		public __children__() {
+			function wrap(node: WNode | VNode) {
+				function w() {
+					return node;
+				}
+
+				Object.keys(node).forEach((key) => ((w as any)[key] = (node as any)[key]));
+
+				return w;
+			}
+
 			if (this._children.some((child) => child.domNode.getAttribute && child.domNode.getAttribute('slot'))) {
 				return [
 					this._children.reduce((slots, child) => {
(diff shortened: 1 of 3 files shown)


matt-gadd:
This is going to be huge 👍

PR #716, opened 2020-04-02
Support rendering multiple routes to an Outlet

### LINKED ISSUE #715: Support rendering multiple routes into a single target outlet
# Enhancement

Change `@dojo/framework/routing` to support rendering multiple routes into the same `outlet`, an `outlet` is considered as a target in an application that depending on the matched route can display varying output.

## The Problems

The concept of an outlet in routing doesn't map to being able to creating static layouts, instead outlets map directly to a unique route. This makes it verbose to use the `Outlet` widget as you have to define one for each route that could be rendered.

```tsx
import { create, tsx } from '@dojo/framework/core/vdom';
import Outlet from '@dojo/framework/routing/Outlet';

const factory = create();

const App = factory(function App() {
    return (
        <div>
            <Outlet id="foo" renderer={(matchDetails) => {
                return <SomeWidget />    
            }} />
            <Outlet id="bar" renderer={(matchDetails) => {
                return <SomeWidget />    
            }} />
            <Outlet id="baz" renderer={(matchDetails) => {
                return <SomeWidget />    
            }} />
            <Outlet id="qux" renderer={(matchDetails) => {
                return <SomeWidget />    
            }} />
        </div>
    );
});
```

## Proposal

Add an additional configuration option to the routing configuration to uniquely identify the route and re-purpose  the `outlet` configuration to more appropriately indicate the section of the application that the route will be rendered.

For an application with a simple layout that includes a side bar and a main view, assuming that the side bar of the application is visible for all routes, with the main layout changing based on the user interaction (route change). For simple applications with few routes, this is fine using the multiple outlet configuration shown above, but as the application grows it can become messy and hard to manage.

```
---------------------------------------------------------
|         |                                             |
|         |                                             |
|         |                                             |
|         |                                             |
|         |                                             |
| sidebar |                  main                       |
|         |                                             |
|         |                                             |
|         |                                             |
|         |                                             |
|         |                                             |
---------------------------------------------------------
```

Changing the concept of an outlet to reference the location in application that should render, helps engineers to easily grok where in the application content is going to be rendered for each route and provides more control within the rendering widget itself.

```tsx
const routeConfig = [
    {
        id: 'foo',
        outlet: 'main',
        path: 'foo'
    },
    {
        id: 'bar',
        outlet: 'main',
        path: 'bar'
    },
    {
        id: 'baz',
        outlet: 'main',
        path: 'baz'
    },
    {
        id: 'qux',
        outlet: 'main',
        path: 'qux'
    }
];
```

With a simple routing configuration shown above all the routes with an `outlet` of `main` would trigger the a single outlet to re-render, providing the required details to determine what view needs to be rendered. 

```tsx
import { create, tsx } from '@dojo/framework/core/vdom';
import Outlet from '@dojo/framework/routing/Outlet';

const factory = create();
const App = factory(function App() {
    return (
        <div>
            <Sidebar />
            <Outlet id="main">
            {{
                foo: () => <Foo />,
                bar: () => <Bar />,
                baz: () => <Baz />,
                qux: (matchDetails) => {
                    return <Qux params={matchDetails.params.id}/>;
                }
             }}
            </Outlet>
        </div>
    );
});
```
By default the keys of the children match to the route ids that are matched for the outlet id, however this could be customised using the `matcher` property to restrict rendering to based on the type of match (exact, partial, etc) or for a label that is not represented by the routes at all and manually determined based on any logic.

```tsx
import { create, tsx } from '@dojo/framework/core/vdom';
import Outlet from '@dojo/framework/routing/Outlet';

const factory = create();
const App = factory(function App() {
    return (
        <div>
            <Sidebar />
            <Outlet id="main" matches={(defaultMatches, matchDetails) => {
                 // do stuff
                 return defaultMatches;
            }}>
            {{
                foo: () => <Foo />,
                bar: () => <Bar />,
                baz: () => <Baz />,
                qux: (matchDetails) => {
                    return <Qux params={matchDetails.params.id}/>;
                }
             }}
            </Outlet>
        </div>
    );
});
```

As well as taking an object for the children and matching being determined by whether there the key matches an active route id or the custom logic determined by the user defined `matcher` property,  a function can passed that will render for _any_ matches or the outlet.

```tsx
import { create, tsx } from '@dojo/framework/core/vdom';
import Outlet from '@dojo/framework/routing/Outlet';

const factory = create();
const App = factory(function App() {
    return (
        <div>
            <Sidebar />
            <Outlet id="main">
               {(matchDetails) => <SomeWidget params={matchDetails.params} />}
           </Outlet>
        </div>
    );
});
```

The existing `Outlet` behaviour would still be available but renamed as `Route`, which accurately describes the behaviour.

```tsx
import { create, tsx } from '@dojo/framework/core/vdom';
import Outlet from '@dojo/framework/routing/Outlet';

const factory = create();

const App = factory(function App() {
    return (
        <div>
            <Route id="foo" renderer={() => {
                return <Foo />
            }}/>
        </div>
    );
});
```

**Note:** The route id and changing `Outlet` to `Route` should be automatically migrated by the cli-upgrade-app tool.

agubler:
**Type:** feature

**Description:**

Changes the concept of an `Outlet` to support rendering multiple matching "routes" to a single outlet. Now using an outlet gives flexibility to render views for an application's routes in a more natural layout, reduces repetitive uses of an `Outlet` that duplicate layout and logic.

The existing `Outlet` behaviour has been moved to `Route` that more accurately describes its behaviour.

Resolves #715

### DIFF
--- a/docs/en/routing/introduction.md
+++ b/docs/en/routing/introduction.md
@@ -275,7 +275,7 @@ To take advantage of the code splitting there are 4 rules:
 1.  The routing configuration needs to be the default export in the `src/routes.ts` module.
 2.  The widgets must be the default export of their module.
 3.  The `renderer` property must be defined inline.
-4.  The `id` and `outlet` in the routing config and must be static and defined inline.
+4.  The `id` and `outlet` in the routing config must be static and defined inline.
 
 > src/routes.ts
 
--- a/docs/en/routing/introduction.md
+++ b/docs/en/routing/introduction.md
@@ -112,7 +112,7 @@ Path parameters are placeholders in the routing configuration that will match an
 ```ts
 export default [
 	{
-		id: 'home,
+		id: 'home',
 		path: 'home/{page}',
 		outlet: 'home'
 	}
--- a/docs/en/routing/introduction.md
+++ b/docs/en/routing/introduction.md
@@ -77,7 +77,7 @@ export default factory(function App() {
 });
 ```
 
-or using outlets and the `Outlet` widget, check out the [`Outlet` documentation](/learn/routing/outlets) for more information;
+or using outlets and the `Outlet` widget, check out the [`Outlet` documentation](/learn/routing/outlets) for more information:
 
 ```tsx
 import { create, tsx } from '@dojo/framework/core/vdom';
--- a/docs/en/routing/introduction.md
+++ b/docs/en/routing/introduction.md
@@ -56,7 +56,7 @@ const r = renderer(() => <App />);
 r.mount({ registry });
 ```
 
--   Add a `Route` widget to show the text "Home" when the `home` route is visited. `Route` is a widget that display something when the path for the route id is matched. The application's `src/routes.ts` file associates a route to an id via the `Route`'s `id` property.
+-   Add a `Route` widget to show the text "Home" when the `home` route is visited. `Route` is a widget that displays something when the path for the route id is matched. The application's `src/routes.ts` file associates a route to an id via the `Route`'s `id` property.
 
 > src/App.tsx
 
@@ -119,7 +119,7 @@ export default [
 ];
 ```
 
-The parameters values are injected into to matching `Route`'s `renderer` property.
+The parameter values are injected into the matching `Route`'s `renderer` property.
 
 > src/App.tsx
 
@@ -214,7 +214,7 @@ export default [
 
 ## Using link widgets
 
-The `Link` widget is a wrapper around an anchor tag that enables consumers to specify an route `id` to create a link to. If the generated link requires specific path or query parameters that are not in the route, they can be passed via the `params` property.
+The `Link` widget is a wrapper around an anchor tag that enables consumers to specify a route `id` to create a link to. If the generated link requires specific path or query parameters that are not in the route, they can be passed via the `params` property.
 
 Link Properties:
 
--- a/docs/en/routing/introduction.md
+++ b/docs/en/routing/introduction.md
@@ -12,15 +12,16 @@ Dojo's Routing package provides a first class declarative routing solution for w
 
 ## Adding routing to an application
 
--   Add an initial route configuration that defines a single url path that maps to an identifier referred to as an `outlet`. Outlets will be described later in the documentation.
+-   Add an initial route configuration that defines a single url path that maps to a route identifier and an `outlet` name.
 
 > src/routes.ts
 
 ```ts
 export default [
 	{
+		id: 'home',
 		path: 'home',
-		outlet: 'home'
+		outlet: 'main'
 	}
 ];
 ```
@@ -45,20 +46,20 @@ const r = renderer(() => <App />);
 r.mount({ registry });
 ```
 
--   Add an `outlet` widget to show the text "Home" when the `home` route is visited. Outlets are widgets that display something when a route is matched. The application's `src/routes.ts` file associates a route to an outlet via the outlet's `id` property.
+-   Add a `Route` widget to show the text "Home" when the `home` route is visited. `Route` is a widget that display something when the path for the route id is matched. The application's `src/routes.ts` file associates a route to an id via the `Route`'s `id` property.
 
 > src/App.tsx
 
 ```tsx
 import { create, tsx } from '@dojo/framework/core/vdom';
-import Outlet from '@dojo/framework/routing/Outlet';
+import Route from '@dojo/framework/routing/Route';
 
 const factory = create();
 
 export default factory(function App() {
 	return (
 		<div>
-			<Outlet id="home" renderer={() => <div>Home</div>} />
+			<Route id="home" renderer={() => <div>Home</div>} />
 		</div>
 	);
 });
@@ -76,26 +77,27 @@ Path parameters are placeholders in the routing configuration that will match an
 ```ts
 export default [
 	{
+		id: 'home,
 		path: 'home/{page}',
 		outlet: 'home'
 	}
 ];
 ```
 
-The parameters values are injected into to matching `Outlets`'s `renderer` property.
+The parameters values are injected into to matching `Route`'s `renderer` property.
 
 > src/App.tsx
 
 ```tsx
 import { create, tsx } from '@dojo/framework/core/vdom';
-import Outlet from '@dojo/framework/routing/Outlet';
+import Route from '@dojo/framework/routing/Route';
 
 const factory = create();
 
 export default factory(function App() {
 	return (
 		<div>
-			<Outlet id="home" renderer={(matchDetails) => <div>{`Home ${matchDetails.params.page}`}</div>} />
+			<Route id="home" renderer={(matchDetails) => <div>{`Home ${matchDetails.params.page}`}</div>} />
 		</div>
 	);
 });
@@ -108,7 +110,8 @@ Query parameters can also be added to route URLs. As with normal query parameter
 ```ts
 export default [
 	{
-		path: 'home/{page}',
+		id: 'home',
+		path: 'home/{page}?{queryOne}&{queryTwo}',
 		outlet: 'home'
 	}
 ];
@@ -118,14 +121,14 @@ export default [
 
 ```tsx
 import { create, tsx } from '@dojo/framework/core/vdom';
-import Outlet from '@dojo/framework/routing/Outlet';
+import Route from '@dojo/framework/routing/Route';
 
 const factory = create();
 
 export default factory(function App() {
 	return (
 		<div>
-			<Outlet
+			<Route
 				id="home"
 				renderer={(matchDetails) => {
 					const { queryParams } = matchDetails;
@@ -137,7 +140,7 @@ export default factory(function App() {
 });
 ```
 
-If the browser is pointed to the URL path `/home/page?queryOne=modern&queryTwo=dojo`, then the query parameters are injected into the matching `Outlet`'s `renderer` method as an object of type `MatchDetails` and accessed via that object's `queryParams` property. Using this URL, the page would show "Hello modern-dojo". If a query parameter is not provided, then its value will be set to `undefined`.
+If the browser is pointed to the URL path `/home/page?queryOne=modern&queryTwo=dojo`, then the query parameters are injected into the matching `Route`'s `renderer` method as an object of type `MatchDetails` and accessed via that object's `queryParams` property. Using this URL, the page would show "Hello modern-dojo". If a query parameter is not provided, then its value will be set to `undefined`.
 
 ## Default route and parameters
 
@@ -148,6 +151,7 @@ If the browser is pointed to the URL path `/home/page?queryOne=modern&queryTwo=d
 ```ts
 export default [
 	{
+		id: 'home',
 		path: 'home',
 		outlet: 'home',
 		defaultRoute: true
@@ -162,6 +166,7 @@ If the default route has path or query parameters a map of defaults need to be s
 ```ts
 export default [
 	{
+		id: 'home',
 		path: 'home/{page}',
 		outlet: 'home',
 		defaultRoute: true,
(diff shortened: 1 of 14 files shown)


docs/en/routing/introduction.md
```diff
@@ -45,10 +56,29 @@ const r = renderer(() => <App />);
 r.mount({ registry });
 ```
 
--   Add an `outlet` widget to show the text "Home" when the `home` route is visited. Outlets are widgets that display something when a route is matched. The application's `src/routes.ts` file associates a route to an outlet via the outlet's `id` property.
+-   Add a `Route` widget to show the text "Home" when the `home` route is visited. `Route` is a widget that display something when the path for the route id is matched. The application's `src/routes.ts` file associates a route to an id via the `Route`'s `id` property.
```
maier49, line 59:
```suggestion
-   Add a `Route` widget to show the text "Home" when the `home` route is visited. `Route` is a widget that displays something when the path for the route id is matched. The application's `src/routes.ts` file associates a route to an id via the `Route`'s `id` property.
```

docs/en/routing/introduction.md
```diff
@@ -76,26 +112,27 @@ Path parameters are placeholders in the routing configuration that will match an
...
 	{
+		id: 'home',
 		path: 'home/{page}',
 		outlet: 'home'
 	}
 ];
 ```
 
-The parameters values are injected into to matching `Outlets`'s `renderer` property.
+The parameters values are injected into to matching `Route`'s `renderer` property.
```
maier49, line 122:
```suggestion
The parameter values are injected into the matching `Route`'s `renderer` property.
```

docs/en/routing/introduction.md
```diff
@@ -174,12 +214,12 @@ export default [
 
 ## Using link widgets
 
-The `Link` widget is a wrapper around an anchor tag that enables consumers to specify an `outlet` to create a link to. If the generated link requires specific path or query parameters that are not in the route, they can be passed via the `params` property.
+The `Link` widget is a wrapper around an anchor tag that enables consumers to specify an route `id` to create a link to. If the generated link requires specific path or query parameters that are not in the route, they can be passed via the `params` property.
```
maier49, line 217:
```suggestion
The `Link` widget is a wrapper around an anchor tag that enables consumers to specify a route `id` to create a link to. If the generated link requires specific path or query parameters that are not in the route, they can be passed via the `params` property.
```

docs/en/routing/introduction.md
```diff
@@ -228,51 +268,56 @@ export default factory(function App() {
...
-When using `@dojo/cli-build-app`, Dojo supports automatic code splitting by default for all top level outlets. This means that all widgets referenced within the `Outlet`s `renderer` will include a specific bundle for the outlet that will be loaded lazily when a user accesses the route.
+When using `@dojo/cli-build-app`, Dojo supports automatic code splitting by default for all top level routes. This means that all widgets referenced within the `Route`s `renderer` will include a specific bundle for the route that will be loaded lazily when a user accesses the route.
 
 To take advantage of the code splitting there are 4 rules:
 
 1.  The routing configuration needs to be the default export in the `src/routes.ts` module.
 2.  The widgets must be the default export of their module.
 3.  The `renderer` property must be defined inline.
-4.  The outlet `id` must be static and defined inline.
+4.  The `id` and `outlet` in the routing config and must be static and defined inline.
```
maier49, line 278: Not sure if "routing config and" is missing the and's "what" or just an extra "and"

docs/en/routing/supplemental.md
```diff
@@ -51,45 +57,189 @@ This example would register the following routes and outlets:
 | `/about/knitting` | `about-services` |
 | `/about/sewing`   | `about-services` |
 
-The `about-services` outlet has been registered to match any path after `/about` This is at odds with the other registered outlets, `about-company` and `about-history`, however Dojo routing ensures that the correct outlet is matched in these scenarios.
+The `about-services` route has been registered to match any path after `/about` This is at odds with the other registered routes, `about-company` and `about-history`, however Dojo routing ensures that the correct routes is matched in these scenarios.
+
+# Outlets
+
+An outlet represents a visual location of an application that renderers different content depending which route has been matched. Using outlets reduces boilerplate required compared to using routes, multiple routes can be associated to the same outlet to more naturally and accurately structure the application output.
```
maier49, line 64:
```suggestion
An outlet represents a visual location of an application that renderers different content depending on which route has been matched. Using outlets reduces boilerplate required compared to using routes, multiple routes can be associated to the same outlet to more naturally and accurately structure the application output.
```

docs/en/routing/supplemental.md
```diff
@@ -51,45 +57,189 @@ This example would register the following routes and outlets:
...
+|  menu  |                   main                     | side-menu |
+|        |                                            |           |
+|        |                                            |           |
+|        |                                            |           |
+|        |                                            |           |
+|        |                                            |           |
+-------------------------------------------------------------------
+```
+
+With a route configuration below that specifies all the main pages to the main content outlet, but the `widget` to a `side-menu` outlet. This enables building an application that constantly renders the main content depending on route, but also include a right hand side menu for all children routes of the `widget` route.
```
maier49, line 84:
```suggestion
The route configuration below specifies all the main pages to the main content outlet, but the `widget` to a `side-menu` outlet. This enables building an application that constantly renders the main content depending on route, but also include a right hand side menu for all children routes of the `widget` route.
```

docs/en/routing/supplemental.md
```diff
@@ -51,45 +57,189 @@ This example would register the following routes and outlets:
...
+				id: 'example'
+				path: 'example/{example}',
+				outlet: 'main'
+			}
+		]
+	}
+];
+```
+
+In the routing configuration above, there are two outlets defined, `main` and `side-menu` and a simplified application layout using outlets is shown below. By default the `Outlet` will render any of the keys that equal a route id that has been match for the outlet, in this case `main`. If a function is passed to the `Outlet` then it will render whenever _any_ route is matched for the outlet specified.
```
maier49, line 119:
```suggestion
In the routing configuration above, there are two outlets defined, `main` and `side-menu`, and a simplified application layout using outlets is shown below. By default the `Outlet` will render any of the keys that equal a route id that has been matched for the outlet, in this case `main`. If a function is passed to the `Outlet` then it will render whenever _any_ route is matched for the outlet specified.
```

docs/en/routing/supplemental.md
```diff
@@ -51,45 +57,189 @@ This example would register the following routes and outlets:
...
+						{({ params: { widget }}) => <SideMenu widget={widget}>}
+					</Outlet>
+				</div>
+			</main>
+		</div>
+	);
+});
+```
+
+The node structure of the `App` looks good and succinctly represents the actual visual output for the user with minimal duplication, there still is a need to duplicate the the usage of the `Example` widget across to different routes. This can be solved by using the `matcher` property to override the default route matching rules. The `matcher` receives the `defaultMatches` and a `matchDetailsMap` in order to make custom matching decisions. In the final example below the usage of `Example` has been combined into a new `key`, `details` that does not exist as a route. This will never match for the outlet unless we override the default matches to set it to true when either the `example` or `overview` route has matched. Finally in the `details` renderer the example property has been defaulted to `overview` to maintain the same behavior as before.
```
maier49, line 159:
```suggestion
The node structure of the `App` looks good and succinctly represents the actual visual output for the user with minimal duplication, there still is a need to duplicate the usage of the `Example` widget across to different routes. This can be solved by using the `matcher` property to override the default route matching rules. The `matcher` receives the `defaultMatches` and a `matchDetailsMap` in order to make custom matching decisions. In the final example below the usage of `Example` has been combined into a new `key`, `details` that does not exist as a route. This will never match for the outlet unless we override the default matches to set it to true when either the `example` or `overview` route has matched. Finally in the `details` renderer the example property has been defaulted to `overview` to maintain the same behavior as before.
```

maier49: I removed a double `the` in that suggestion. The diff is useless

docs/en/routing/supplemental.md
```diff
@@ -51,45 +57,189 @@ This example would register the following routes and outlets:
...
+			</main>
+		</div>
+	);
+});
+```
 
 # Router API
 
-The Dojo Router exposes an API that can be used to generate and navigate to links, get the params for the current route and check if an outlet id has been matched.
+The Dojo Router exposes an API that can be used to generate and navigate to links, get the params for the current route and check if an route id has been matched.
```
maier49, line 203:
```suggestion
The Dojo Router exposes an API that can be used to generate and navigate to links, get the params for the current route and check if a route id has been matched.
```

docs/en/routing/supplemental.md
```diff
@@ -51,45 +57,189 @@ This example would register the following routes and outlets:
...
+-   `link(route: string, params: Params = {}): string | undefined`: Generate a link based on the route id and optionally params. If no params are passed it will attempt to use the current routes parameters, then any default parameters provided in the routing configuration. If a link cannot be generated, `undefined` is returned.
 -   `setPath(path: string): void`: Sets the path in the router.
 -   `get currentParams(): { [string: index]: string }`: Returns parameters in the current route
--   `getOutlet(outletIdentifier: string): OutletContext | undefined`: Returns the `OutletContext` for an outlet id if it is currently matched. If the outlet id is not matched, then return `undefined`.
+-   `getRoute(id: string): RouteContext | undefined`: Returns the `RouteContext` for an route id if it is currently matched. If the route id is not matched, then return `undefined`.
```
maier49, line 208:
```suggestion
-   `getRoute(id: string): RouteContext | undefined`: Returns the `RouteContext` for a route id if it is currently matched. If the route id is not matched, then return `undefined`.
```

docs/en/routing/supplemental.md
```diff
@@ -253,25 +408,25 @@ export default [
 ];
 ```
 
--   given the above route definition, if the URL path is set to `/#home/about`, then `isExact()` will evaluate to `false` for the `Outlet` with the id "home" and `true` for the an `Outlet` that is a child of the home `Outlet` with the id "about" as shown in the following file:
+-   given the above route definition, if the URL path is set to `/#home/about`, then `isExact()` will evaluate to `false` for the `Route` with the id "home" and `true` for the an `Route` that is a child of the home `Route` with the id "about" as shown in the following file:
```
maier49, line 411:
```suggestion
-   given the above route definition, if the URL path is set to `/#home/about`, then `isExact()` will evaluate to `false` for the `Route` with the id "home" and `true` for a `Route` that is a child of the home `Route` with the id "about" as shown in the following file:
```

docs/en/routing/supplemental.md
```diff
@@ -498,21 +659,21 @@ r.mount({ registry });
 
 These history managers work like adapters, meaning that custom history managers can be implemented by fulfilling the history manager interface.
 
-# Error outlet
+# Error route
 
-A special `outlet` called `errorOutlet` is registered for that will match when the route doesn't match (`exact` or `partial`) any outlet in the routing configuration. You can use this `outlet` to render a widget to inform the user that the route does not exist.
+A special `route` called `errorRoute` is registered for that will match when the route doesn't match (`exact` or `partial`) any route in the routing configuration. You can use this `route` to render a widget to inform the user that the route does not exist.
```
maier49, line 664:
```suggestion
A special `route` called `errorRoute` is registered that will match when the route doesn't match (`exact` or `partial`) any route in the routing configuration. You can use this `route` to render a widget to inform the user that the route does not exist.
```

docs/en/routing/introduction.md
```diff
@@ -45,10 +56,29 @@ const r = renderer(() => <App />);
...
+		<div>
+			<Route id="home" renderer={() => <div>Home</div>} />
+			<Route id="about" renderer={() => <div>About</div>} />
+			<Route id="profile" renderer={() => <div>Profile</div>} />
+		</div>
+	);
+});
+```
+
+or using outlets and the `Outlet` widget, check out the [`Outlet` documentation](/learn/routing/outlets) for more information;
```
dylans, line 80: incomplete thought / start of sentence?

docs/en/routing/introduction.md
dylans, line 278:
```suggestion
4.  The `id` and `outlet` in the routing config must be static and defined inline.
```

docs/en/routing/introduction.md
agubler: @dylans I think this is supposed to be a colon for the example below

agubler:
@dylans I've updated based on your review.

PR #717, opened 2020-04-06
Spread widget properties in vdom

agubler:
**Type:** bug

**Description:**

Not spreading the properties in vdom when copying them to the wrappers properties can allow users to inadvertently mutate the property object so the diff fails and widgets are not updated as expected.

### DIFF
--- a/tests/core/unit/vdom.tsx
+++ b/tests/core/unit/vdom.tsx
@@ -3487,6 +3487,35 @@ jsdomDescribe('vdom', () => {
 				});
 			});
 
+			it('should close properties', () => {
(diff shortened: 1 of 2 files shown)


tests/core/unit/vdom.tsx
```diff
@@ -3487,6 +3487,35 @@ jsdomDescribe('vdom', () => {
 				});
 			});
 
+			it('should close properties', () => {
```
matt-gadd, line 3490: clone?

agubler: updated, thanks :D

PR #719, opened 2020-04-07
Upgrade to TS 3.8

### LINKED ISSUE #718: Add support for TypeScript 3.8.x and potentially 3.9.x
**Enhancement** 
We should update framework to make sure it works with current `latest` which is 3.8.3. We should also consider testing with the `next` version of 3.9.x which should be shipping soon.

maier49:
**Type:** feature

**Description:**

Resolves #718

### DIFF
--- a/package.json
+++ b/package.json
@@ -86,7 +86,7 @@
     "rimraf": "~2.6.2",
     "selenium-webdriver": "3.6.0",
     "sinon": "~4.1.3",
-    "typescript": "3.8.3"
+    "typescript": "3.4.5"
   },
   "lint-staged": {
     "*.{ts,tsx,md}": [
--- a/package.json
+++ b/package.json
@@ -86,7 +86,7 @@
     "rimraf": "~2.6.2",
     "selenium-webdriver": "3.6.0",
     "sinon": "~4.1.3",
-    "typescript": "3.4.5"
+    "typescript": "3.8.3"
   },
   "lint-staged": {
     "*.{ts,tsx,md}": [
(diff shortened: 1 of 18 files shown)


src/stores/process.ts
```diff
@@ -321,7 +320,8 @@ export function processExecutor<T = any, P extends object = DefaultPayload>(
 					});
 
 					if (isThenable(result)) {
-						return result.then((result) => {
+						result.then();
```
agubler, line 323: Is this supposed to be here?

package.json
```diff
@@ -86,7 +86,7 @@
     "rimraf": "~2.6.2",
     "selenium-webdriver": "3.6.0",
     "sinon": "~4.1.3",
-    "typescript": "3.4.5"
+    "typescript": "3.8.3"
```
agubler, line 89: Do we want to change the version, we are mostly trying to make the package forwards compatible. Is that possible?

src/stores/process.ts
maier49: Nope, removed

package.json
maier49: Oh that was just my misunderstanding. I've reverted the package version, still builds fine against 3.4.5

maier49:
This worked with the `next` version of TypeScript as well with no further changes needed than for 3.8

maier49:
@agubler Ok so now if we see a file using `registerRouterInjector` and none of the imports will match the route regex (determined by resolving the import relative to the directory of the current file) then it logs out a warning.

PR #750, opened 2020-04-23
Adding JSON deserialization to custom element properties (#748)

### LINKED ISSUE #748: Support JSON serialization in custom elements
Currently, custom elements convert a widget's properties into attributes, properties, and events.

- *attributes* - Any string property
- *events* - Any function property that starts with "on"
- *properties* - Everything!  All widget properties get exposed as regular properties to some extent (events are still a bit special).

Sometimes, a widget property might take a configuration object.

    <CheckboxGroup options={[{ value: 'cat' }, { value: 'dog' }, { value: 'fish' }]}>
      {{
        label: 'pets'
      }}
    </CheckboxGroup>

During the custom element transformation, `options` would be defined as a property, since it is not a string type.  This makes setting the property a little more difficult than it needs to be.

    <dojo-checkbox-group id="checkbox">
        <label slot="label">pets</label>
    </dojo-checkbox-group>
    ...
    <script>
    checkbox.options = [{value:'cat'},{value:'dog'},{value:'fish'}];
    </script>

Ideally, any non-primitive (string, number, bool), non-function type would be created as an attribute with automatic JSON serialization.

CheckboxGroup would then look like,

    <dojo-checkbox-group options="[{value:'cat'},{value:'dog'},{value:'fish'}]">
        <label slot="label">pets</label>
    </dojo-checkbox-group>

rorticus:
**Type:** feature

**Description:**

Adding JSON serialization to custom element properties. Any property that is not an attribute will have it's new value deserialized as JSON. If the result is valid, it will be set as the property value.  If the result is not valid, it will be ignored.

Resolves #748

### DIFF
--- a/src/core/registerCustomElement.ts
+++ b/src/core/registerCustomElement.ts
@@ -107,10 +107,15 @@ function registerThemeInjector(theme: any, themeRegistry: Registry): ThemeInject
 }
 
 export function create(descriptor: any, WidgetConstructor: any): any {
-	const { attributes = [], registryFactory = () => new Registry() } = descriptor;
+	const { attributes = [], properties = [], registryFactory = () => new Registry() } = descriptor;
 	const attributeMap: any = {};
 
 	attributes.forEach((propertyName: string) => {
+		const attributeName = propertyName.toLowerCase();
+		attributeMap[attributeName] = propertyName;
+	});
+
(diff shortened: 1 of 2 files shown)


agubler:
@rorticus could you add some details in the custom element docs?

agubler:
The failure is unrelated and has been fixed on master

PR #760, opened 2020-04-28
Use theme injector payload if no theme property available

### LINKED ISSUE #758: Theme variant class not available immediately after setting theme
**Bug** 

Package Version: current

**Code**

```ts
if (!theme.get()) {
	theme.set(materialTheme, 'dark');
}
const variantClass = theme.variant();
// variantClass is undefined
```

**Expected behavior:**

The variant class should be available immediately

**Actual behavior:**

The variant class is undefined.

**Proposed Solution**
The theme middleware should get the theme payload from the them injector rather than relying on the diffProperty function to put it in properties as in the immediate scenario, it has not yet been ran.

tomdye:
**Type:** bug / feature

**Description:**

Uses theme injector payload for variant class when theme property not yet set / influenced by diffproperty.

Resolves #758

### DIFF
--- a/src/core/middleware/theme.ts
+++ b/src/core/middleware/theme.ts
@@ -141,6 +141,14 @@ export const theme = factory(
 				const { theme } = properties();
 				if (theme && isThemeWithVariant(theme)) {
 					return theme.variant.value.root;
+				} else {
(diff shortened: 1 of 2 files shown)


tomdye:
Closing as replaced by https://github.com/dojo/framework/pull/768

PR #766, opened 2020-04-30
Wildcard routing

### LINKED ISSUE #330: Route config support wildcard
**Enhancement** 

Package Version: 5.0

I want design a REST API `{owner}/{project}/tree/{branch}/*` to view the folder structure.

```text
|- root
    |- folder1
          |- folder2
                 |- folder3
                        |- file1
```

1. use `{owner}/{project}/tree/{branch}` to get `folder1`
1. use `{owner}/{project}/tree/{branch}/folder1` to get `folder2`
1. use `{owner}/{project}/tree/{branch}/folder1/folder2` to get `folder3`

It is similar to github's repo url.

**Code**

I use router config like below first

```json
{
    "path": "{owner}/{project}/tree/{branch}/{folderPathes}",
    "outlet": "view-project-tree"
}
```

`folderPathes` can be empty or null, 'folder1', 'folder1/folder2' which contains `/`

But the router can not switch '{owner}/{project}/tree/{branch}/folder1' to '{owner}/{project}/tree/{branch}/folder1/folder2'

may be here do a check

https://github.com/dojo/framework/blob/master/src/routing/Router.ts#L244

```ts
if (route.segments[segmentIndex] === undefined) {
    type = 'partial';
    break;
}
```
because `route.segments` length is greater than `params` length

So I expect to support wildcard like this

```json
{
    "path": "{owner}/{project}/tree/{branch}/*",
    "outlet": "view-project-tree"
}
```

**Expected behavior:**

```json
{
    "path": "tree/*",
    "outlet": "view-tree"
}
```

`*` can be zero or more segments

1. if view the root folder, then the path is `tree`, not include the ending `/`
1. if view the second level, then the path is `tree/folder1`
1. if view the third level, then the path is `tree/folder1/folder2`
1. ...

maier49:
**Type:** feature

**Description:**
A proposed implementation of wildcard routing

* A wildcard route is indicated by `*`. It will match a route normally up until the asterisk and then match anything after that.
* A slightly higher penalty is imposed on the route `score` than for a route param, so that a more specific route will always win.
* Introduces the `wildcard` route type and `wildcardSegments` to the `RouteContext` and `MatchDetails`. This will include all the segments of the path matched by the asterisk.
* Adds additional checking for `wildcard` routes so that a change in route segments triggers a route change, even if the same wildcard route is still matched.

Resolves #330

### DIFF
--- a/docs/en/routing/introduction.md
+++ b/docs/en/routing/introduction.md
@@ -212,6 +212,53 @@ export default [
 ];
 ```
 
+## Wildcard Routes
+
+The `*` character can be used to indicate a wildcard route. The route will be matched normally up until the `*` and will match
+any path at that point. A wildcard route will never be preferred over another matching route without a wildcard. The `*` implicitly indicates the end of the
+match, and any segments specified after the `*` in the route config will be ignored. Any additional segments in the actual URL will be passed with
+the `matchDetails` in an array property called `wildcardSegments`.
+
+```ts
+export default [
+	{
+		id: 'catchall',
+		// Anything after the asterisk will be ignored in this config
+		path: '*',
+		outlet: 'catchall'
+	},
+	// This path will be preferred to the wildcard as long as it matches
+	{
+		id: 'home',
+		path: 'home',
+		outlet: 'home'
+	}
+];
+```
+
+All segments after and including the matched `*` will be injected into the matching `Route`'s `renderer` property as `wildcardSegments`.
+
+> src/App.tsx
+
+```tsx
+import { create, tsx } from '@dojo/framework/core/vdom';
+import Route from '@dojo/framework/routing/Route';
+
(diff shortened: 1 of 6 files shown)


agubler:
@maier49 what would the behaviour be here for these two route configurations? Would `foo/bar/1/2/3/4/5/qux` match `"foo/bar/*/qux"` over the catch all `"foo/bar/*"`?

```ts
{
    path: "foo/bar/*/qux" 
},
{
    path: "foo/bar/*" 
}
```

agubler:
I think these changes might need to be updated with master also

maier49:
@agubler With the implementation here it would  just match whichever one is checked first. The wildcard section is explicitly treated as the final section of the route, anything past that is ignored.

PR #768, opened 2020-05-01
Do not return global injected theme as the theme property

### LINKED ISSUE #758: Theme variant class not available immediately after setting theme

agubler:
**Type:** bug

**Description:**

Changes to stop the `theme` middleware using `diffProperty` to augment a widgets `theme` property with a globally configured theme. Currently the `theme` middleware assumes that the `properties().theme` value is always correct (either the explicitly passed theme or the globally injected theme), which is problematic if the `theme.set(` api is called during a render (for example initially setting the theme). 

The `theme` property should only need to be used to pass onto a child widget (to ensure they use an explicitly passed theme if one is passed to the parent), otherwise only the middleware APIs should be used to work with themes. Internally the `theme` middleware has been changed to pick the theme to use by checking the widget's theme property first and falling back to the globally configured theme.

Resolves #758

### DIFF
--- a/src/core/middleware/theme.ts
+++ b/src/core/middleware/theme.ts
@@ -41,20 +41,10 @@ export const theme = factory(
 		let themeKeys = new Set();
 
 		diffProperty('theme', properties, (current, next) => {
-			const themeInjector = injector.get<ThemeInjector>(INJECTED_THEME_KEY);
 			const diffResult = auto(current.theme, next.theme);
 			if (diffResult.changed) {
 				icache.clear();
 				invalidator();
-			}
-
-			if (!next.theme && themeInjector) {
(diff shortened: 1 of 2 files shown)


agubler:
Supersedes #760

PR #769, opened 2020-05-02
[WIP] Improve typings for data transform option

agubler:
**Type:** bug

**Description:**

Overhauls the `data` and `resource` typings to improve the ergonomics for passing a `transform` that describes how to convert the resource data to the data middleware data. The `transform` is now passed to the `resource` factory, along with the optional `data`, if the resource item structure matches the data item structure then the transform is not required. When the two data types do not match then a transform is required when using the resource.

When passing a `resource` to a widget, it is required to call the factory, a bare resource can no longer be passed without being called.

```tsx
// This will now error
<MyWidget resource={resource} />

// This is fine as long as the resource/data item structures match
<MyWidget resource={resource()} />

// Using a transform to convert the resource data to the data item structure
<MyWidget resource={resource({ transform: { value: 'resourceProp' )} />
```

* `transform` is optional based on whether the resource can fulfil the data structure.
* `transforms` only supports single keys, not possible to provide composite keys as part of the transform
* `resources` the default page number is now set to `1`
* `defaultFilter` no longer checks for `value` and `label` and uses the passed query directly
* A default `data` middleware is no longer exported, only the data middleware factory
* Re-write the introduction resource documentation

**TODO**: Update the supplemental documentation

### DIFF
--- a/src/core/middleware/data.ts
+++ b/src/core/middleware/data.ts
@@ -1,6 +1,7 @@
 import { create, invalidator, destroy, diffProperty } from '../vdom';
 import { Resource, ResourceOptions, ResourceQuery, TransformConfig } from '../resource';
 import { Invalidator } from '../interfaces';
+import { auto } from '../diff';
 
 type Query = { [key: string]: string | undefined };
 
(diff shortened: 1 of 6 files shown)


src/core/middleware/data.ts
```diff
@@ -163,42 +159,38 @@ export function createDataMiddleware<T = void>() {
 
 		diffProperty('resource', properties, ({ resource: currentResource }, { resource: nextResource }) => {
 			if (isResourceWithData(currentResource) && isResourceWithData(nextResource)) {
-				if (
-					nextResource.data !== currentResource.data ||
-					nextResource.data.length !== currentResource.data.length
-				) {
-					resourceWithDataMap.delete(nextResource.data);
+				const result = auto(currentResource.data, nextResource.data);
```
agubler, line 162: Use the `auto` diff to detect if the data has changed.

src/core/resource.ts
```diff
@@ -48,29 +56,21 @@ export interface DataTemplate<S = {}> {
...
+	return response.then !== undefined;
 }
 
-export function defaultFilter(query: ResourceQuery[], item: any) {
-	let filterValue = '';
-
-	query.forEach((q) => {
-		if (q.keys.indexOf('value') !== -1) {
-			filterValue = q.value || '';
+export function defaultFilter(queries: ResourceQuery[], item: any) {
```
agubler, line 63: change the default filter to be generic and work on the query passed... the previous version was specific to the usage when it was created in widgets.

src/core/resource.ts
```diff
@@ -223,7 +223,7 @@ export function createResource<S>(config: DataTemplate<S> = createMemoryTemplate
 	}
 
 	function getOrRead(options: ResourceOptions): S[] | undefined {
-		const { pageNumber, query, pageSize } = options;
+		const { pageNumber = 1, query, pageSize } = options;
```
agubler, line 226: Defaults the pageNumber to `1` (which it probably always should be if there is not one set)

tomdye: 👍

src/core/resource.ts
```diff
@@ -8,10 +8,16 @@ export interface ResourceOptions {
 	pageSize?: number;
 }
 
-export type ResourceQuery = { keys: string[]; value: string | undefined };
-
-export interface Resource<S = any> {
-	(data: S[]): { resource: Resource<S>; data: S[] };
+export type ResourceQuery = { keys: string; value: string | undefined };
```
tomdye, line 11: Should this change now? it's no longer `keys`. This could just be a simple object now

agubler: I should change the property name, I don't it being an object here because it means the user doesn't have to `Object.keys`

PR #773, opened 2020-05-11
feat: drag middleware

### LINKED ISSUE #514: Drag Middleware

bitpshr:
**Type:** feature

**Description:**

This pull request adds a new module that exposes drag middleware.

Resolves #514 
References https://github.com/dojo/widgets/issues/1429

### DIFF
--- a/src/core/middleware/drag.ts
+++ b/src/core/middleware/drag.ts
@@ -1,0 +1,240 @@
+import global from '../../shim/global';
+import { assign } from '../../shim/object';
+import { deepAssign } from '../util';
+import { create, destroy, invalidator, node } from '../vdom';
+import icache from './icache';
+
+/**
+ * Results from a drag operation
+ */
+export interface DragResults {
+	/**
+	 * The movement of pointer during the duration of the drag state
+	 */
+	delta: Position;
+
+	/**
+	 * Is the DOM node currently in a drag state
+	 */
+	isDragging: boolean;
+
+	/**
+	 * A matrix of posistions that represent the start position for the current drag interaction
+	 */
+	start?: PositionMatrix;
+}
+
+/**
+ * An x/y position structure
+ */
+export interface Position {
+	/** Horizontal coordinate */
+	x: number;
+
+	/** Vertical coordinate */
+	y: number;
+}
+
+/**
+ * A matrix of x/y positions
+ */
+export interface PositionMatrix {
+	/**
+	 * Client x/y position
+	 */
+	client: Position;
+
+	/**
+	 * Offset x/y position
+	 */
+	offset: Position;
+
+	/**
+	 * Page x/y position
+	 */
+	page: Position;
+
+	/**
+	 * Screen x/y position
+	 */
+	screen: Position;
+}
+
+interface NodeData {
+	dragResults: DragResults;
+	invalidate: () => void;
+	last: PositionMatrix;
+	start: PositionMatrix;
+}
+
+function getDelta(start: PositionMatrix, current: PositionMatrix): Position {
+	return {
+		x: current.client.x - start.client.x,
+		y: current.client.y - start.client.y
+	};
+}
+
+function createNodeData(invalidate: () => void): NodeData {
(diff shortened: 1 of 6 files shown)


src/core/middleware/drag.ts
```diff
@@ -0,0 +1,240 @@
...
+		state.dragResults.delta = getDelta(state.start, state.last);
+		state.dragResults.isDragging = false;
+		state.invalidate();
+		icache.set('dragging', undefined);
+
+		event.preventDefault();
+		event.stopPropagation();
+	}
+
+	if (!icache.get('initialized')) {
```
agubler, line 203: As `initialized` is always `undefined`, i'm not sure what this check does?

agubler: Also this differs from the existing `Drag` meta, it will add a new handler for every widget that uses the `drag` middleware. In the a`Drag` meta this was done by having a singleton controller.

src/core/middleware/drag.ts
```diff
@@ -0,0 +1,240 @@
...
+
+	/**
+	 * Screen x/y position
+	 */
+	screen: Position;
+}
+
+interface NodeData {
+	dragResults: DragResults;
+	invalidate: () => void;
```
agubler, line 65: What do we need to store a widgets invalidate function for?

src/core/middleware/drag.ts
```diff
@@ -0,0 +1,240 @@
...
+const factory = create({ destroy, icache, invalidator, node });
+
+export const drag = factory(({ middleware: { destroy, icache, invalidator, node } }) => {
+	const nodeMap = icache.getOrSet('nodeMap', () => new WeakMap<HTMLElement, NodeData>());
+
+	function getData(target: HTMLElement): { state: NodeData; target: HTMLElement } | undefined {
+		if (nodeMap.has(target)) {
+			return { state: nodeMap.get(target)!, target };
+		}
+		if (target.parentElement) {
```
agubler, line 139: What is the reason for referencing `target.parentElement`, we shouldn't be potentially bleeding out of the widget from a middleware?

src/core/middleware/drag.ts
bitpshr: Good catch; it looks like we can either kill this `initialized` check altogether, always creating new handlers, or we can move the listeners to the module scope.

bitpshr: I ended up removing this initialization check, creating listeners for each middleware use, with tests in-place to verify the listener cleanup

bitpshr:
Thanks @agubler, ready for another pass when there's a chance.

src/core/middleware/drag.ts
```diff
@@ -0,0 +1,233 @@
...
+		win.removeEventListener('pointerdown', onDragStart);
+		win.removeEventListener('pointermove', onDrag, true);
+		win.removeEventListener('pointerup', onDragStop, true);
+	});
+
+	return {
+		get(key: string | number): Readonly<DragResults> {
+			const domNode = node.get(key);
+
+			if (!domNode) return emptyResults;
```
agubler, line 214: generally use `{}` even for one line returns

src/core/middleware/drag.ts
```diff
@@ -0,0 +1,233 @@
...
+	isDragging: false
+});
+
+const factory = create({ destroy, icache, invalidator, node });
+
+export const drag = factory(({ middleware: { destroy, icache, invalidator, node } }) => {
+	const nodeMap = icache.getOrSet('nodeMap', () => new WeakMap<HTMLElement, NodeData>());
+
+	function getData(target: HTMLElement): { state: NodeData; target: HTMLElement } | undefined {
+		if (nodeMap.has(target)) {
```
agubler, line 134: If you did 
```tsx
const state = nodeMap.get(target);
if (state) {
    return { state, target };
}
```

then you wouldn't need the `!`

PR #797, opened 2020-05-27
Resource verbiage changes

jellison:
A series of docs changes to correct grammar and make the content more readable.

Includes one sample code fix as well.

### DIFF
--- a/docs/en/resources/supplemental.md
+++ b/docs/en/resources/supplemental.md
@@ -1,6 +1,6 @@
 # Resource Concepts
 
-Dojo resources is designed to provide a cohesive and consistent mechanism for working with data within a Dojo application. These are the core concepts for Dojo Resources:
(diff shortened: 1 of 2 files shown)


docs/en/resources/supplemental.md
```diff
@@ -1,16 +1,16 @@
 # Resource Concepts
 
-Dojo resources is designed to provide a cohesive and consistent mechanism for working with data within a Dojo application. There are core concepts for Dojo Resources:
+Dojo resources is designed to provide a cohesive and consistent mechanism for working with data within a Dojo application. These are the core concepts for Dojo Resources:
```
agubler, line 3: I think this was originally supposed to say `There are three core concepts for Dojo Resources" - happy to go with which ever you think it better.

jellison: Gotcha. That makes sense too.

PR #802, opened 2020-06-11
Expand types for insert after/before

### LINKED ISSUE #801: Incorrect type for `children` parameter of `.insertBefore` (and variants)
**Bug**

In the new assertion API, the `children` parameter (i.e., 2nd param) of `insertBefore`, and `insertAfter` are not typed appropriately.

**Actual**

https://codesandbox.io/s/gracious-microservice-wui7d?file=/src/main.tsx

**Expected**

The expected type of the `children` parameter matches the defined type of the widget's `children`.

**Workaround**

Casting to any bypasses the type check and works as expected: https://codesandbox.io/s/beautiful-gates-owu3t?file=/src/main.tsx

maier49:
**Type:** bug

**Description:**
This expands the types for `insertBefore/insertAfter`. This would I believe be a backwards compatible fix but I think may be the appropriate fix in general.
 We currently allow inserting before/after at the root and when the parent is not wrapped, and requiring a parent would break both of those use cases. Making the parent argument optional could allow the user to opt into typing but seems klugey since the parent isn't actually needed.

Additionally the only concern with widening the type is that we'd allow the parent's child type to be violated, but even if we enforced it here the `replace` function accepts any DNode. It seems inconsistent if our API enforces the types of siblings but not the node itself. 
Resolves #801

### DIFF
--- a/src/testing/renderer.ts
+++ b/src/testing/renderer.ts
@@ -138,20 +138,14 @@ export interface AssertionResult {
 	): AssertionResult;
 	insertBefore<T extends WidgetBaseInterface>(
 		target: Wrapped<Constructor<T>>,
-		children: TemplateChildren<T['children']>
+		children: TemplateChildren
 	): AssertionResult;
-	insertBefore<T extends WidgetFactory>(
-		target: Wrapped<T>,
-		children: TemplateChildren<T['children']>
-	): AssertionResult;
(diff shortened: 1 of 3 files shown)


agubler:
@maier49 I think this is backwards compatible and fixes the ability to insert siblings - at the moment it types children to the wrapped siblings child type which isn’t correct at all. The fix would allow the children type of the parent to be violated like you mention though.

I think the correct fix for the next major release would be to require the parent on the API to be able to correctly enforce the child type when working with siblings.

PR #815, opened 2020-07-15
Custom element resource support

maier49:
**Type:** feature

**Description:**
Adds a flag to factory's using the resource middleware

* Uses attributes/properties on the custom element to configure the resource (right now only works if those are not used but does not reserve them)
* Defaults to a memory template and empty data set
* Does not expose `createTemplate` functions since they only provide typing, instead taking a template directly

Example usage [here](https://github.com/maier49/custom-element-resource/blob/master/src/index.html#L73)  
Resolves dojo/widgets#1481


maier49:
@matt-gadd The prototype now handles everything in the middleware (and has tests). It also occurs to me that with this change to the logic we could also change the type on the `resource` property to allow passing an object with all the options directly instead of needing a parent that uses the resource middleware to create a resource prop

PR #822, opened 2020-07-21
Fix mock middleware types

maier49:
**Type:** bug

**Description:**
`Exclude<keyof X, 'Y'>` returns a unioned string type not the object type. Normally using `Omit` or `Pick` would fix that but In the case of the MiddlewareResultFactory the type is `never` becuase the call signature isn't a key.

This changes the mocked middleware type to the call signature itself.

### DIFF
--- a/tests/testing/unit/harness/harness.tsx
+++ b/tests/testing/unit/harness/harness.tsx
@@ -1,5 +1,3 @@
-import createICacheMock from '../../../../src/testing/mocks/middleware/icache';
-
 const { describe, it } = intern.getInterface('bdd');
 const { assert } = intern.getPlugin('chai');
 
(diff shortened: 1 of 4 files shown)


tests/testing/unit/harness/harness.tsx
```diff
@@ -620,6 +622,42 @@ describe('harness', () => {
 			));
 		});
 
+		it('should mock icache middleware', () => {
```
matt-gadd, line 625: should this test be in the icache mock test suite over the harness?

maier49: I should maybe just change the name, or use fake middleware? It's really just to test that it allows a mock at all, not to verify the icache mock functionality

Issue #770, opened 2020-05-04
Lazy load custom elements on demand


rorticus:
It would be great if custom elements could be loaded on demand after they are used for the first time.  All the user would need to include was the bootstrap file, and the actual elements will be loaded on their own.

rorticus:
Made a POC [here](https://github.com/rorticus/lazy-element-test/tree/master).  I think it proves the concept, but requires a bit of clean up.

- Make sure nothing broke with class based widgets
- Generated bundle names are numbers, which isn't great

agubler:
This has been landed throughout the packages

PR #825, opened 2020-08-13
Custom element updates for dynamic loading

maier49:
**Type:** feature

**Description:**
Framework changes as described in #770


maier49:
@rorticus Awaiting a synchronous value is fine though right? And the `_readyCallback` is already called asynchronously so I'm not sure where the problem would be there. The issues with backwards compatibility would be always expecting a factory function instead of the `WidgetConstructor` itself like we used to. But I'm not sure there's a way around that (and it should be fixed by just rebuilding unless it was being used manually).

matt-gadd (approved):

PR #840, opened 2020-09-09
Handle ignored children in decorate

### LINKED ISSUE #839: Test renderer: can't call property for conditional wrapped node
**Bug** 

`r.property(WrappedNode, 'propName')` does not work for a node that is not expected in the next assertion

Package Version: 7.0.3

**Code**

(See [`FileUploadInput` test](https://github.com/msssk/widgets/blob/0ec69738e2eb946da84b7e4c79010cad7d094765/src/file-upload-input/tests/unit/FileUploadInput.spec.tsx#L152-L185))

```typescript
	it('handles dragenter, dragleave, and the overlay', function() {
		const r = renderer(function() {
			return <FileUploadInput />;
		});
		const WrappedOverlay = wrap('div');

		r.expect(baseAssertion);
		r.property(WrappedRoot, 'ondragenter', { preventDefault: noop });

		r.expect(
			baseAssertion
				.setProperty(WrappedRoot, 'classes', [
					null,
					fixedCss.root,
					css.root,
					css.dndActive,
					false
				])
				.append(WrappedRoot, function() {
					return [
						<WrappedOverlay
							key="overlay"
							classes={[fixedCss.dndOverlay, css.dndOverlay, false]}
							ondragleave={noop}
						/>
					];
				})
		);

		r.property(WrappedOverlay, 'ondragleave');
		r.expect(baseAssertion);
	});
```

**Expected behavior:**

`ondragleave` should be triggered on `WrappedOverlay`

**Actual behavior:**

`ondragleave` is not triggered

maier49:
**Type:** bug

**Description:**
Allows finding wrapped nodes when the actual array has additional children that are undefined or booleans.

This fix resolves the failure that led to #839 being created, but the actual bug was not what we originally thought.
Resolves #839

### DIFF
--- a/tests/testing/unit/renderer.tsx
+++ b/tests/testing/unit/renderer.tsx
@@ -14,6 +14,23 @@ const noop: any = () => {};
 
 class ChildWidget extends WidgetBase<{ id: string; func?: () => void }> {}
 
+const ConditionalRender = create({ icache })(({ middleware: { icache } }) => {
+	return v('div', {}, [
+		icache.get('render')
+			? v('div', {
+					onclick: () => {
+						icache.set('render', false);
+					}
+			  })
+			: null,
+		v('div', {
+			onclick: () => {
+				icache.set('render', true);
+			}
+		})
+	]);
+});
+
 class MyWidget extends WidgetBase {
 	_count = 0;
 	_result = 'result';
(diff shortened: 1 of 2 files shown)


tests/testing/unit/renderer.tsx
```diff
@@ -222,6 +239,19 @@ describe('test renderer', () => {
 			);
 		});
 
+		it('trigger property when there are undefined children in actual render', () => {
```
msssk, line 242: would "triggers property" be better?

tests/testing/unit/renderer.tsx
```diff
@@ -222,6 +239,19 @@ describe('test renderer', () => {
...
+			const WrappedRoot = wrap('div');
+			const WrappedChild = wrap('div');
+			const WrappedConditional = wrap('div');
+			const baseTemplate = assertion(() => w(WrappedRoot, {}, [w(WrappedChild, { onclick: noop })]));
```
agubler, line 246: If you're using wrapped vnodes with the functional API we should be using `v()` and then calling `.tag` from the WrappedNode - `v(WrappedRoot.tag` - alternatively using TSX. I thought there was mention of this in the docs but I can't find it.

tests/testing/unit/renderer.tsx
maier49: Yeah that sounds better

matt-gadd (approved):

PR #847, opened 2020-09-25
Adds Event Handler Options

### LINKED ISSUE #810: Passive event listeners
**Enhancement**

Chrome has taken to logging a warning if you add a listener to scroll events.  
```
[Violation] Added non-passive event listener to a scroll-blocking 'mousewheel' event. Consider marking event handler as 'passive' to make the page more responsive.
```

These warnings can be prevented by indicating that the listener is passive and will not call `event.preventDefault()` by passing `{ passive: true }` when the event listener is added. Similarly, passing `{ passive: false }` indicates that the listener is not passive, and should disable the warning.

**Code**
```
<div
  onwheel={(event) => { ...anything... }}
/>
```

samends:
**Type:** Feature

Adds ability to use event handler options as described here: https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener. The syntax would be the following:

```
const scrollElementDiv = v('div', { onscroll: {
	passive: true,
	handler: (event: UIEvent) => {
		// handle onscroll event here
	}
} });
```
As describe in the open issue, this would allow events on elements to be label as `passive`, preventing the warning on chrome. This will also allow a developer to use the options `capture` and `once`.

Resolves #810

### DIFF
--- a/src/core/vdom.ts
+++ b/src/core/vdom.ts
@@ -73,8 +73,7 @@ import {
 	TimeAttributes,
 	TrackAttributes,
 	VideoAttributes,
-	EventOptions,
-	EventWithOptions
+	EventOptions
 } from './interfaces';
 import { Registry, isWidget, isWidgetBaseConstructor, isWidgetFunction, isWNodeFactory } from './Registry';
 import { auto } from './diff';
@@ -1548,10 +1547,6 @@ export function renderer(renderer: () => RenderResult): Renderer {
 		}
 	}
 
-	function isInstanceOfEventWithOptions(value: any): value is EventWithOptions {
-		return value;
-	}
-
 	function setProperties(
 		domNode: HTMLElement,
 		currentProperties: VNodeProperties = {},
@@ -1616,11 +1611,14 @@ export function renderer(renderer: () => RenderResult): Renderer {
 					setValue(domNode, propValue, previousValue);
 				} else if (propName !== 'key' && propValue !== previousValue) {
 					const type = typeof propValue;
-					if (isInstanceOfEventWithOptions(propValue)) {
+					if (type === 'object' && propValue && 'handler' in propValue) {
 						const { handler, ...eventOptions } = propValue;
 						updateEvent(domNode, propName.substr(2), handler, previousValue, eventOptions);
-					}
-					if (type === 'function' && propName.lastIndexOf('on', 0) === 0 && includesEventsAndAttributes) {
+					} else if (
+						type === 'function' &&
+						propName.lastIndexOf('on', 0) === 0 &&
+						includesEventsAndAttributes
+					) {
 						updateEvent(domNode, propName.substr(2), propValue, previousValue);
 					} else if (type === 'string' && propName !== 'innerHTML' && includesEventsAndAttributes) {
 						updateAttribute(domNode, propName, propValue, nextWrapper.namespace);
--- a/src/core/vdom.ts
+++ b/src/core/vdom.ts
@@ -131,7 +131,44 @@ export namespace tsx.JSX {
 		time: TimeAttributes;
 		track: TrackAttributes;
 		video: VideoAttributes;
-		[key: string]: VNodeProperties;
+		[key: string]:
+			| VNodeProperties
+			| CanvasAttributes
+			| ColAttributes
+			| DetailsAttributes
+			| DialogAttributes
+			| EmbedAttributes
+			| FieldsetAttributes
+			| FormAttributes
+			| IFrameAttributes
+			| ImgAttributes
+			| ButtonAttributes
+			| InputAttributes
+			| LabelAttributes
+			| LinkAttributes
+			| MapAttributes
+			| MenuAttributes
+			| MeterAttributes
+			| MetaAttributes
+			| AnchorAttributes
+			| ObjectAttributes
+			| OlAttributes
+			| OptionAttributes
+			| OutputAttributes
+			| ParamAttributes
+			| ProgressAttributes
+			| QuoteAttributes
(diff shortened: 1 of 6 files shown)


matt-gadd:
@samends I don't think the current implementation correctly works with `once` because of the way event handlers are re-attached. for example try this:

```tsx
import { renderer, create, tsx, invalidator } from "@dojo/framework/core/vdom";

const factory = create({ invalidator });

const App = factory(function App({ middleware: { invalidator } }) {
  return (
    <div>
      <button
        onclick={{
          once: true,
          handler: () => {
            console.log("hello world");
            invalidator();
          }
        }}
      >
        Click
      </button>
    </div>
  );
});

const r = renderer(() => <App />);
r.mount();
````

File: tests/testing/unit/harness/harness.tsx [hunk 1]
```diff
@@ -301,7 +301,11 @@ describe('harness', () => {
 			const h = harness(() => w(MyWidget, {}));
 			h.trigger('*[key="span"]', (node: WNode | VNode) => {
 				if (isVNode(node)) {
-					return node.properties.onclick;
+					const onclick = node.properties.onclick;
```
matt-gadd, line 304: How come we had to change this here?

matt-gadd:
Does this change work with the test renderer?

src/core/vdom.ts
```diff
@@ -1609,7 +1648,14 @@ export function renderer(renderer: () => RenderResult): Renderer {
 					setValue(domNode, propValue, previousValue);
 				} else if (propName !== 'key' && propValue !== previousValue) {
 					const type = typeof propValue;
-					if (type === 'function' && propName.lastIndexOf('on', 0) === 0 && includesEventsAndAttributes) {
+					if (type === 'object' && propValue && 'handler' in propValue) {
```
agubler, line 1651: Would this also need the `includesEventsAndAttributes` to ensure that events don't get registered twice? Another thing would be that we probably need to add support for this new style of event handlers when using `dom()` function (https://github.com/dojo/framework/blob/master/src/core/interfaces.d.ts#L171)

When support has been added for that then it might make sense to put all the logic handling either a function or object with a handler inside the `updateEvent` function.

samends:
Hmm @matt-gadd maybe we shouldn't offer `once` as an option then

tests/testing/unit/harness/harness.tsx
samends: This now needs to check if the event is an object with `handler` or an function

matt-gadd: this is a test though isn't it, and you haven't changed the test? so why has the vnode changed?

matt-gadd: I think I now that this is a type guard, and you can't possibly ever get to the handler code - so I think this would be better as just a type cast instead

matt-gadd: Hmm so the more I think about this I think the more problematic this type change is for users downstream. Given how this is not the common use case, have we considered not changing `VNodeProperties` at all, and instead creating a utility function that basically creates a function with properties decorated on it (thus adhering to the existing interface?). so something like:

```tsx
<button onclick={ eventWithOptions({ capture: true }, () => console.log('onclick')) }>Button</button>
```

or perhaps even nicer if we just want to solve the `passive` case for now like the issue raised:

```tsx
<button onscroll={ passive(() => console.log('onscroll')) } >Button</button>
```

samends: Hmm I haven't considered something like that but it would make the typing alot easier

samends: Let me look into this

src/core/vdom.ts
```diff
@@ -1296,7 +1335,7 @@ export function renderer(renderer: () => RenderResult): Renderer {
 			};
 		}
 
-		domNode.addEventListener(eventName, callback);
+		domNode.addEventListener(eventName, callback, options);
```
matt-gadd, line 1338: I think you will have to do a check here to support IE11, as it doesn't take an options bag so passing anything truthy to it will mean it does a capture.

samends: So IE doesn't seem to have an equivalent here, could we still implement this with a note saying IE doesn't support event options?

matt-gadd: Yeah none of the options bag is implemented in IE other than capture and it’s a Boolean as the third parameter I believe.

agubler: Yup that’s correct, it doesn’t error passing an object bag but it does consider it “truthy” so capture is enabled. We’ll need to sniff IE11 and and not pass the options bag at all I think. That can be documented that the “passive” helper isn’t supported in IE11

tests/testing/unit/harness/harness.tsx
samends: For this to work you would need to change the types so that the function could optionally take any function

samends: Going back to this problem what your suggesting would still require a change to the event types so we'd still need to check to see if there's a handler here. What could work instead is if we optionally took two values for events?

agubler: @samends I'm still not sure what the problem with a wrapper function is with regards to types? It looks like you could do something like this without requiring any changes to types (as it works for v7), and then hook into the property on the handler in the vdom to deal with it?

https://codesandbox.io/s/wonderful-sound-kijth?file=/src/main.tsx

matt-gadd:
Closing as resolved in https://github.com/dojo/framework/pull/856

PR #856, opened 2020-10-23
Adds support for event proxying and event options

### LINKED ISSUE #810: Passive event listeners

matt-gadd:
**Type:** feature

**Description:**
This adds support for 2 features:

***Event Proxying*** 

Before this change, events were removed and re-attached if the callback changed across renders. Unfortunately this is a very common occurrence due to inline functions encouraged. With this change, a wrapping proxy is used which simply points to the callback, which means we do not have to re-attach the event.

***Event Options***

Added a new special typed property called `oneventoptions` which currently supports `passive`; an array of supported eventlisteners. This allows you to set whether an event is passive or not.(https://developers.google.com/web/updates/2016/06/passive-event-listeners).

Squatting on a property on a `VNodeProperties` is not ideal, but it is at least prefixed with the existing `on` squat. And this method is the least intrusive to the existing behaviour.

Example usage:
```tsx
<div 
  onscroll={ () => { /* whatever your event callback is */ } } 
  oneventoptions={ { passive: ['onscroll'] /* sets the above event to be passive */ } }
>Scrollable Content</div>
```

Event options are not supported on `addEventListener` at all in older browsers (IE) so a sniff has been added to `has` called `dom-passive-event`.

* Upgraded jsdom to support passive events
* Upgraded webcomponents polyfill for tests due to some strange behaviour/interaction with events in IE

Resolves #810


src/core/vdom.ts
```diff
@@ -1312,9 +1337,13 @@ export function renderer(renderer: () => RenderResult): Renderer {
 			const isEvent = propName.substr(0, 2) === 'on' || onlyEvents;
 			const eventName = onlyEvents ? propName : propName.substr(2);
 			if (isEvent && !properties[propName]) {
-				const eventCallback = _eventMap.get(previousProperties[propName]);
-				if (eventCallback) {
-					domNode.removeEventListener(eventName, eventCallback);
+				const proxyEvents = _eventMap.get(domNode) || {};
+				let proxyEvent: { proxy: EventListener; callback: Function; options: any } | undefined =
```
agubler, line 1341: Could this be `EventMapValue | undefined`?

src/core/vdom.ts
```diff
@@ -1236,6 +1236,8 @@ function wrapFunctionProperties(id: string, properties: any) {
 	return props;
 }
 
+type EventMapValue = { proxy: EventListener; callback: Function; options: any } | undefined;
```
agubler, line 1239: Could we type options? `passive: boolean` or similar?

agubler (approved):
Just a couple of type suggestions, otherwise :+1:

src/core/vdom.ts
matt-gadd: I wanted to actually use `EventListenerOptions` but I think the type didn't have passive. I'll type it though.

src/core/vdom.ts
matt-gadd: yeah it can actually just be inferred, i just forgot to remove it.

PR #866, opened 2020-11-27
Refines the create template function APIs

### LINKED ISSUE #864: Consider merging createResourceTemplate, createResourceTemplateWithInit and createMemoryResource in Resources
**Enhancement**
It would be nice if these were all merged into one so we have a single way to do things, with the memory template just being the default if no object is passed containing `read`, `find` etc

agubler:
**Type:** feature

**Description:**

Merges createResourceTemplate, createResourceTemplateWithInit and createMemoryResourceTemplate in a single `createResourceTemplate` function that supports creating templates with and without init options and defaults to the old "memory template" which is not going to be called the "default template".

Resolves #864


agubler:
@matt-gadd is this the kind of change you were thinking for the issue? This still requires docs to be updated and might be nice to provide a shim for the removed APIs to make it non breaking with a deprecated warning?

matt-gadd (approved):

matt-gadd (approved):

PR #869, opened 2020-12-03
Support replace for the router and history managers

agubler:
**Type:** bug

**Description:**

For `redirect` the new route should replace the current one so that the browser navigation behaves as expected. To support a `replace` feature has been added to both the router and the history managers.

### DIFF
--- a/tests/routing/unit/Router.ts
+++ b/tests/routing/unit/Router.ts
@@ -678,7 +678,6 @@ describe('Router', () => {
 
 	describe('redirect', () => {
 		it('should redirect from a default route', () => {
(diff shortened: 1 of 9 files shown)


tests/routing/unit/Router.ts
```diff
@@ -678,7 +678,6 @@ describe('Router', () => {
 
 	describe('redirect', () => {
 		it('should redirect from a default route', () => {
-			debugger;
```
agubler, line 681: Whoops

matt-gadd (approved):

PR #872, opened 2020-12-08
Update insertSiblings type definition

### LINKED ISSUE #870: insertSiblings not overloaded
**Bug**

The `insertSiblings` function does not have an interface for using with Wrapped nodes as `insertBefore` and `insertAfter` do: https://github.com/dojo/framework/blob/master/src/testing/renderer.ts#L174

Package Version: 7.0.5

samends:
**Type:** Bug

**Description:**

Resolves #870 
Updates type definitions for `insertSiblings` function


agubler:
@samends I think we can add a test that uses this API with a function based widget - without your changes it should fail with compilation errors and after should work?

samends:
Make sense, I'll add that in

agubler (approved):

PR #893, opened 2021-03-26
Function to compare nodes to widgets

agubler:
**Type:** feature

**Description:**

Functions to determine whether the a node is a type of widget.

### DIFF
--- a/src/core/vdom.ts
+++ b/src/core/vdom.ts
@@ -72,7 +72,8 @@ import {
 	ThAttributes,
 	TimeAttributes,
 	TrackAttributes,
-	VideoAttributes
+	VideoAttributes,
+	WNodeFactoryTypes
 } from './interfaces';
 import { Registry, isWidget, isWidgetBaseConstructor, isWidgetFunction, isWNodeFactory } from './Registry';
 import { auto } from './diff';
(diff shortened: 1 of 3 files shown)


src/core/vdom.ts
```diff
@@ -525,6 +525,13 @@ export function isElementNode(value: any): value is Element {
 	return !!value.tagName;
 }
 
+type WidgetFactory = WNodeFactory<any> | OptionalWNodeFactory<any> | DefaultChildrenWNodeFactory<any>;
+
+export function typeOf(node: WNode, factory: WidgetFactory) {
```
matt-gadd, line 530: can you not typeguard this?

matt-gadd (changes requested):

src/core/vdom.ts
agubler: We can add `: node is WNode` but in order to narrow the properties seems like it's going to be a larger change due to the default of WNode being any, i.e. it remains as `any` - It's possible to support this but it's a much wider typing change that is required.

agubler: I’m going to spend some time to understand all the changes that would be required. Ultimately it would be better if it was possible to type guard the node type. So worth investing more time 👍

agubler: @matt-gadd Initially I thought it would require changing the default generics of `WNode` which had knock on effects across multiple modules, however it was actually only the typing of the nodes returns from tsx that needed to be typed to `unknown` so that narrowing could work - the nodes created by `w` and the function factory worked already. I've pushed the changes please let me know if there's anything else.

PR #912, opened 2021-07-28
Example of simple css file, and more complicated third party library …

### LINKED ISSUE #911: Improve 'Static Assets' doc page with examples.
**Enhancement**

Improve 'Static Assets' doc page with examples.

Package Version: master 

**Code**

Documentation update.

**Expected behavior:**

Should be able to create and use files in `assets` folder.

**Actual behavior:**

Path issues caused grief because it wasn't specified to use an absolute path.

neekfenwick:
…in the assets folder.

**Type:** feature

**Description:**

Resolves #911

### DIFF
--- a/docs/en/building/supplemental.md
+++ b/docs/en/building/supplemental.md
@@ -149,31 +149,16 @@ Static assets can be added to an `assets/` directory at the project root. At bui
 
 The build also parses `src/index.html` for CSS, JavaScript, and image assets, hashes them and includes them in the output directory. A favicon can be added to `src` and referenced by `src/index.html`. The build will then hash the file and copy it to the output directory with a file name of `favicon.[hash].ico`.
 
-## Examples
-### Simple .css File
-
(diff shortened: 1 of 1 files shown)


agubler:
Thanks for the contribution, I think we should change this to only provide an example of how to reference files in the assets directory instead of specific scenarios.

```html
<link rel="shortcut icon" href="/assets/favicons/favicon.ico">
```

neekfenwick:
I've simplified the content based on your feedback.  The reason for the details was that I'd not understood some nuances of how the process worked, for example my first example of how to use /assets was to put a simple.css file in there and reference it from index.html, which I learned was an antipattern because you can just put static assets like that in /src and the build process will hash and include them.  This is mentioned in the existing content but wasn't really clear to me until I'd seen it working.  Maybe because I'm not familiar with webpack and more modern web build frameworks.

In fact the existing content states "A favicon can be added to `src` and referenced by `src/index.html`." which seems to go against your suggest of using a favicon as an example of using `/assets`.

I'm left a little confused as to what you'd accept :)

agubler (approved):

